PR #73, opened 2026-07-17
fix(test): pin isolation off in the bind-failure Start test

DonaldMurillo:
TestStart_BindFailureRunsShutdown occupies a port and asserts `Start` returns an error. In a linked git worktree, the default `worktree` isolation mode remaps the occupied port to a free one — so `Start` binds successfully and serves forever, failing the test with a misleading "Start did not return" in **every worktree checkout** while passing on a plain clone (which is why CI never saw it).

The test pins the bind-failure contract, not isolation behavior, so it now sets `GOFASTR_ISOLATION=off` for its duration. Verified 3/3 green in a linked worktree where it previously failed 100%, and still green on a plain checkout.

### DIFF
--- a/framework/start_bindfail_test.go
+++ b/framework/start_bindfail_test.go
@@ -15,6 +15,14 @@ import (
 // context — instead of returning early and leaking every worker, cron, and
 // queue an earlier start phase spawned.
 func TestStart_BindFailureRunsShutdown(t *testing.T) {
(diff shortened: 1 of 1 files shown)


DonaldMurillo:
Superseded by #74, which includes this exact commit (cherry-picked) as part of the v0.29.0 batch. Closing to keep one PR.

PR #128, opened 2026-07-23
v0.41.0 — readiness and UI contracts: BFF posture, entity sub-configs, typed fields, runtime module split

DonaldMurillo:
Rebased, review-hardened, and split re-land of the `ci/pre-push-skip-deletions` branch (its hook fix already merged separately; the 86-file mega-commit is now 12 themed commits on top of v0.40.0).

## What ships

- **`auth.WithBFFPosture`** — cookie-only browser auth in one option: `__Host-` session cookie, exact-origin allowlist + credentialed CORS on the API prefix, bearer-JWT rejection (gfsk_ tokens pass), global CSRF with only the self-guarding logout route exempt. (`feat(auth)`)
- **Grouped `EntityConfig` sub-configs** — `Scope`/`Pagination`/`Exposure` pointer groups, declaration + blueprint mirrors, flat fields deprecated through a compatibility window. (`feat(entity)`, `feat(cli)`)
- **Typed form fields** — `ui.TextField`/`NumberField`/`DateField`. (`feat(ui)`)
- **Widget runtime module split** — demand-loaded widgethelpers/widgetfocus/widgetlinks; every gzip-budget override deleted, core back under 12KB. (`feat(runtime)`)
- **Both-order entity/screen collision diagnostics**, `gofastr audit lint` precision pass, `init` version pinning, `public_openapi` blueprint key, composed-page a11y fixes on the example site, fail-closed error handling in idempotency/outbox, stability policy + external pilot program docs, changelog for v0.41.0.

## Fixes found in review (all TDD'd on this branch)

1. **Grouped `Exposure` was ignored by App route/MCP wiring** — `TryEntity`/`GroupEntity` read the raw config instead of the normalized `e.Config`, so `Exposure{CRUD:&false}` still mounted CRUD routes and the MCP/CRUD contradiction went undetected. Fixed + `framework/exposure_test.go`.
2. **Runtime module self-registration gap** — the extracted helper modules registered scanners without setting `loadedModules`, so the rescan loops never re-wired remounted widgets or poll-swapped rows (kiln chat panel lost draft-restore/charcount/submit-on-enter on reopen; tick-elapsed counters froze). Fixed; the three previously-failing kiln browser tests pass.
3. **`runtime-contract.md` out of sync** with the new sidebar attributes — the embedded-doc sync test failed on the original branch tip; fixed in the same commit as the attribute docs.

## Verification

- `./scripts/test-all.sh` green (initial run's two failures were the upgrades.yml `through` bump — fixed — and a local-only stale `.gofastr/axe-coverage.json` gap-fail, green from clean state; CI always starts clean).
- `./scripts/coverage-floors.sh` all floors hold; gofmt clean; no binaries.
- Four scoped review agents over the full diff: auth/middleware and CLI areas zero findings; the three findings above were the only confirmed defects.


DonaldMurillo:
Multi-model review round (GLM + Sol) complete — 9 verified findings, all fixed in the 7 commits ending 49ff5395:

**Sol (P1s):**
1. Default 404/405 responses bypassed the router middleware chain → CORS preflights died at 405 on OPTIONS-less CRUD paths. Fixed in core/router (defaults now chain-wrapped like custom handlers).
2. `WithBFFPosture` without `WithAPIPrefix` guarded a phantom `/api` while entities mounted bare → guard now fail-closed covers the whole app.
3. `Sec-Fetch-Site: same-site` skipped the Origin comparison → sibling-subdomain forced logout. same-site now falls through to the Origin-host check.
4. SQL lint's taint-name gate dropped quote-adjacent interpolation with ordinary names → quoted interpolation flags unconditionally.
5. Commented `name="_csrf"` counted as CSRF protection → raw-field counting moved to comment-stripped source.
6. shortcut.js accumulated permanent per-element document listeners across widget remounts → rewritten as a single live-DOM dispatcher.
7. Deep-link URL could be pushed after the widget was dismissed mid-module-load → push now checks mounted state (dismiss-side ms-wide window documented as accepted).
8. (P2) Collapsed sidebar labels lost accessible names (display:none on focusable links' only label) → visually-hidden clip pattern, pinned by test.

**GLM:**
9. SQL lint's `INSERT INTO` anchor missed `INSERT OR IGNORE/REPLACE INTO` / `INSERT IGNORE INTO` → anchor accepts the variant forms.

Also fixed en route: the examples HMR readiness test now isolates its axe-coverage root (the shared repo-root manifest made a parallel package's recordings flip strict from absence-warns to gap-fails, scheduling-dependent), and the sidebar e2e probe asserts the clip contract. Full local suite, runtime budgets (untouched — paid for the new guard with dedupe), coverage floors, and both example lint runs are green.

PR #162, opened 2026-08-01
Go 1.27 adoption: toolchain, stdlib uuid, CutLast, goroutine-leak surface, synctest

DonaldMurillo:
Adopts Go 1.27 across the framework, built and gated against go1.27rc2 (stable lands mid-August; flip the go.mod directive to 1.27.0 then — nothing else should need to change).

## What's here, by commit

- **Toolchain + modernizers** — go.mod (and the check-embed testdata module) to `go 1.27rc2`; the new `go fix` analyzers touched 565 files (net −556 lines), behavior-changing omitzero rewrites excluded. `encoding/json` is now backed by json/v2 in the stdlib: faster Unmarshal for free, and the full suite passed unchanged under the rc before any of this work started (baseline run).
- **stdlib `uuid`** — request IDs and crud entity-key defaults stop hand-rolling RFC 4122 bytes. The 32-hex ID generators in outbox/webhook/queue keep their format; those are persisted identifiers.
- **`strings.CutLast`** — six LastIndex-slicing sites converted; five left alone where index arithmetic is the clearer tool.
- **Header-value cap** — Go 1.27 caps request header values at 500 by default; a posture test now fails if anything overrides `MaxHeaderValueCount`.
- **Goroutine-leak surface** — the runtime's new `goroutineleak` profile exposed three ways: `goroutineLeaks` in `/.debug/stats`, `/.debug/goroutineleak` (auth-gated, stacks), and an `app_goroutine_leaks` MCP introspection tool. A regression test pins the app start/serve/stop cycle at zero new leaks. Docs + agent-skill rosters updated (the guidance test enforces this).
- **Shutdown flake, root-caused** — Go 1.27's Body.Close drains and pools connections it used to kill, so a queued Transport dial can complete unused; the server parks it in StateNew, and Server.Shutdown spares StateNew conns for 5s (go#22682) — exactly the test helpers' shutdown deadline. Failed 2-in-5 full-package runs under load, 8/8 after flushing the client pool in the helpers. `App.Shutdown` errors now name the failing phase, which is what made this diagnosable.
- **synctest conversions** — in-memory timing tests in `core/stream` (35.7s → 13.3s; the 10s websocket write-timeout test now finishes instantly) and `battery/queue` (8.5s → 7.0s, converted tests all 0.00s). DB/Redis/socket tests stay real-time; a bubble panics on non-durable blocking.
- **Docs** — changelog entry, Go floor to 1.27 in README/tutorial/deploy/site quickstart, Dockerfile images to golang:1.27.

## Gates

- `./scripts/test-all.sh`: exit 0, 181 packages ok
- Coverage floors hold (run on every commit by the hook)
- Isolated: `core-ui/runtime` (131s), `core-ui/widget`, site `TestAxe_AllPagesAreClean` — all ok
- govulncheck: clean (note: govulncheck must itself be built with the 1.27 toolchain — `GOTOOLCHAIN=go1.27rc2 go install golang.org/x/vuln/cmd/govulncheck@master` — or its type-checker chokes on generic methods in the new stdlib)

## Deliberately not in this PR

- Generic-methods API adoption — wants a real design pass, not a sweep
- Deeper json/v2 API usage in hot paths — the free win is already in
- The remaining ~1s websocket tests (each burns the real 1s peer-close timeout; bubble-safe, mechanical follow-up)
- Merging before Go 1.27 stable — hold until the directive can say `1.27.0`


DonaldMurillo:
Superseded by #212, closing.

This branch was 245 commits behind main, so the Go 1.27 work was re-derived on a current base rather than merged. Every commit here has a 1:1 counterpart there — the toolchain move and modernizers, stdlib `uuid`, `strings.CutLast`, the goroutine-leak profile, the synctest bubbles, the header-value cap, and the keep-alive shutdown fix — with `go1.27rc2` replaced by stable.

One commit was deliberately not carried over: `30dd79a4` ("bootstrap stable Go and let the go.mod directive pick the toolchain"). It swapped `go-version-file: go.mod` for `go-version: stable` only because setup-go could not resolve `1.27rc2`. That constraint is gone on stable, and taking it would have reverted the per-SHA cache work from #193 that landed after this branch was cut.

Nothing else here is unique. #212 carries the work forward.

PR #204, opened 2026-08-17
test(framework): de-flake the seed-wiring poll deadline

DonaldMurillo:
The v0.65.0 release gate went red on `TestAppStart_WiresRunSeeds` — "seeded row not present after App.Start; got 0 rows" at exactly 3.00s — on the merge commit, while the identical tree was green on the PR head and green again on rerun. The test polls a 3-second wall-clock deadline for Start → migrate → seed under a CI box running the full parallel job matrix; that budget is the flake.

The poll breaks the moment the row lands, so raising the deadline to 30s costs nothing when healthy and only extends the report time on a genuine failure.

## Summary by CodeRabbit

* **Tests**
  * Improved startup seeding test reliability by allowing more time for initialization to complete.

### DIFF
--- a/framework/seed_test.go
+++ b/framework/seed_test.go
@@ -515,7 +515,11 @@ func TestAppStart_WiresRunSeeds(t *testing.T) {
 	startErr := make(chan error, 1)
 	go func() { startErr <- app.Start(":0") }()
 
-	deadline := time.Now().Add(3 * time.Second)
+	// Generous deadline: the poll breaks the moment the row lands, so
+	// the budget is only ever spent on failure — 3s was a real 1-in-N
+	// CI flake when Start→migrate→seed ran on a box under full
+	// parallel-job load (v0.65.0's release gate caught exactly that).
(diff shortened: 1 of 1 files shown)


DonaldMurillo:
Shipped in v0.68.0 via #221, which folded this branch in — commit `3a81849f` is on main.

Worth recording why it needed a follow-up: the de-flake here raised the poll deadline from 3s to 30s, and CI failed `TestAppStart_WiresRunSeeds` again on #221 at exactly 30.01s. Raising the deadline made the failure slower, not rarer. The test opened `:memory:`, where modernc gives every pooled connection its own empty database, so a poll that lands on a second connection can never succeed — it just spends the whole budget. It also discarded `app.Start`'s error, which is where `run seeds: ...` would have named the cause instead of leaving "got 0 rows" as the only evidence.

Both fixed in `34f0a7db`: a file-backed database, and the error checked first. Closing as delivered.

PR #205, opened 2026-08-17
v0.66.0: queue lease fencing, blueprint boot gate, and the assessment backlog

DonaldMurillo:
Fixes every verified finding from a clean-room assessment of v0.65.0 (five
independent analysts plus a strategy panel; each finding re-verified by hand
before it was acted on).

## Why this is a minor and not a patch

One fix is breaking. `Queue.Ack`/`Nack` take a `Job` instead of a job ID
string, because the string form carries no claim identity and therefore
cannot fence a stale lease holder. Keeping it would mean keeping an API
whose behaviour is "sometimes silently loses your job". Nothing in this
repo outside `battery/queue` calls it, `gofastr-plugins` does not use the
queue, and pkg.go.dev reports no importers, so the real-world cost is
close to zero — but it cannot honestly be labelled a patch.

## The two that mattered most

**The flagship blueprint generated an app that would not boot.**
`examples/blog/gofastr.yml` seeded `post_id: 1` against UUID string primary
keys. The app compiled, then died at startup with `seed comments: validation
failed` — and that message was unactionable because the generated seed
discarded `ValidationError.Fields()`, where `post_id: ["must be a string
reference"]` lived. It survived because the gate ladder stopped at
compilation: `TestExampleBlueprintsLoad` parses every blueprint and
`TestExampleBlueprintsGenerateAndCompile` compiles each one, but nothing ran
one. Meanwhile the committed `examples/blog/main.go` is hand-written and
seeds via raw `db.Exec`, so it worked while its blueprint rotted beside it.

`TestExampleBlueprintsBoot` adds the missing rung, and caught a second
defect on its first run: `blueprintScreenMountStmt` chained `WithTitle` but
not `WithDescription` for policy-gated screens, so every screen with a
declared description tripped strict mode and meridian's generated app could
not start.

**Both durable queues lost jobs in ordinary crash sequences,** and neither
failure was visible from `Stats` or `Replay`. Reproductions and root causes
are in the commit message for 6d7702a7.

## Everything else

Multipart uploads over 1 MiB (a 1 MiB reader wrapped the body before the
content type was checked); cursor pagination re-serving rows whose sort key
held zero-width characters; repeated `?field_in=` keys dropping all but the
first; nested `?rel.field_in=` with no cap while three sibling paths capped
at 1000; `?page=` near `MaxInt64` wrapping negative through a guard that
existed but was never called; two users overwriting each other by uploading
the same filename.

Eight documentation claims contradicted the code. The worst,
`multi-tenant.md`, taught reading the tenant from a request header —
`TenantMiddleware` ignores any client header precisely because trusting one
lets a caller impersonate a tenant, so the page documented a vulnerability
the code refuses to have.

Two gates ran without gating. `TestEveryPackageIsClassified` ran `go list
./...` from its own package directory, so it checked 1 package instead of
219; verified by mutation that an unclassified tree passed before and fails
now. The doc-example compile gate covered 2 of 445 fenced Go blocks;
enabling 83 more surfaced ten real breaks, including `new(0)` for a
`*time.Duration` and `Post` given an `http.HandlerFunc`.

## Deliberately not done

- **YAML flow mappings still refused.** af5f56f3 records the minimal-subset
  parser as an architectural choice; the diagnostic is the fix. The change
  `core/yaml` would need is named in the branch notes.
- **`TestT2CursorDuplicateKeys` left failing.** `cursor-pagination.md:107`
  documents it: a non-unique single cursor field stalls on ties, use
  `CursorFields` for a tiebreak. Fixing it would change documented semantics.
- **chromedp and sqlmock stay in the module graph.** `cmd/gofastr` imports
  chromedp for `audit a11y`, so removing them means splitting the CLI into
  its own module. A scaffold binary links 19 deps and neither of these, so
  the cost is `go.sum` size, not runtime.

## Verification

`go build`, `go vet`, `gofmt`, and `./scripts/coverage-floors.sh` all clean
against Postgres on :5433 — no floor was lowered; `framework/crud` rose to
97.0 against a 96.9 floor. `./scripts/test-all.sh` green. `examples/site`
re-run isolated after a chromedp-under-load cascade: 821s, zero failures.
Each commit vets standalone.

## Needs a maintainer action this PR cannot take

`historical upgrade fixtures (blocking)` is required by
`scripts/release-required-checks.txt` but is missing from branch
protection's ten contexts, so a PR can merge with it red and the break only
surfaces at release:

```
gh api --method POST \
  repos/DonaldMurillo/gofastr/branches/main/protection/required_status_checks/contexts \
  -f 'contexts[]=historical upgrade fixtures (blocking)'
```

## Summary by CodeRabbit

* **New Features**
  * Added safer queue processing with stale-claim protection and final-attempt dead-lettering.
  * Added unique upload storage keys and a 64 MiB multipart upload limit.
  * Improved repeated `_in` filters, cursor handling, and overflow-safe pagination.
  * Added comprehensive blueprint seed validation and clearer error reporting.
* **Bug Fixes**
  * Prevented upload filename collisions, pagination errors, cursor data loss, and blueprint boot failures.
  * Corrected generated home-page titles and documentation URLs.
* **Documentation**
  * Expanded compile checks across Go examples and updated guidance for uploads, security, tenants, queues, and feature flags.
* **Tests**
  * Added broad coverage for queue recovery, uploads, pagination, blueprints, and example application boot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,109 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
 
 ## [Unreleased]
 
+## [0.66.0] - 2026-08-17
+
+### Fixed
+
+- **BREAKING: a stale queue worker could delete another worker's job.**
+  `Queue.Ack` and `Queue.Nack` now take a `Job` instead of a job ID
+  string, and `RedisQueue` verifies `Job.ClaimToken` before acting. The
+  old signature carried no claim identity, so this sequence lost a job
+  outright: worker A claims job J, A stalls past its lease, worker B
+  re-claims J, A wakes and Acks. A's Ack deleted the processing entry
+  that belonged to B, and when B then crashed the job was on no list, in
+  no hash, and had never been dead-lettered. A mismatched token is now a
+  no-op counted by `StaleClaimCount()`. Update call sites to pass the
+  `Job` returned by `Dequeue`.
+- **A worker crash on a job's final attempt stranded it forever.**
+  `DBQueue` reclaimed expired leases only while `attempts <
+  max_attempts`, but `Dequeue` had already incremented `attempts` — so a
+  crash on the last permitted attempt left the row `claimed` with
+  nothing able to see it, including `Replay`. An expired lease on the
+  final attempt is now dead-lettered and logged.
+- **Multipart uploads over 1 MiB always failed.** The CRUD write
+  handlers wrapped every request body in a 1 MiB `MaxBytesReader` before
+  deciding whether it was multipart, so `ParseMultipartForm` read
+  through a cap 32× smaller than the 32 MiB it was asked for. JSON
+  bodies still cap at `MaxJSONBodyBytes`; multipart bodies now cap at
+  the new `crud.MaxMultipartBodyBytes` (64 MiB) and a request over it
+  gets a 413 instead of a 400.
+- **Cursor pagination re-served rows whose sort key held invisible
+  characters.** `EncodeCursor` kept the raw keyset value while
+  `DecodeCursor` stripped zero-width and bidi codepoints, so paging
+  resumed before the row it had just emitted. Values now decode
+  verbatim; field names, which reach SQL as identifiers, are still
+  scrubbed.
+- **Repeating a `?field_in=` key dropped every occurrence but the
+  first.** `?tag_in=a,b&tag_in=c` filtered on `a,b` alone and said
+  nothing. All occurrences now union, and the 1000-entry cap counts the
+  union. The same cap now applies to relation-scoped lists
+  (`?author.name_in=`), which had no cap at all.
+- **`?page=` near `MaxInt64` wrapped to a negative offset.** The guard
+  in `framework/pagination` existed but nothing called it. It is now
+  wired into the buffered list, the streaming list, and the admin table.
+- **Two users uploading the same filename overwrote each other.** The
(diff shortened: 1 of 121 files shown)


DonaldMurillo:
@coderabbitai review

Four commits since the last pass address all 13 inline findings plus the outside-diff one. Two were resolved by disagreeing, with evidence:

- **RNG failure on `UniqueFilename`**: `crypto/rand.Read` in this Go version "never returns an error, and always fills b entirely" and crashes the program if the OS source fails, so the branch being guarded is unreachable. Rather than thread an error through `UniqueFilename` and the pre-existing exported `file.GenerateFilePath` (a second breaking change for an impossible condition), the dead branch is deleted.
- **`bigint`/`double`/`money`/`numeric` in the seed type check**: `entity.parseFieldType` accepts only `int|integer`, `float|number`, `decimal`, so `type: bigint` fails at generate time with `unknown type "bigint"` — verified by generating one. Widening the seed validator to those aliases would make it accept types the runtime rejects. Reasoning recorded in `docs/agent-notes.md`.

Also went past the two findings as scoped: the `EntityConfig.OwnerField` naming was swept by invariant rather than in the one file flagged, which found four instances of a field that does not exist (it moved under `Scope` in v0.54.0), including in `CLAUDE.md`; and the two same-class allocate-then-reject sites noted as "worth a next-round look" (`include.go` pipe lists, `predicate.go` where-body lists) are fixed here.

Issue #211, opened 2026-08-19
Deferred review findings from PR #206


DonaldMurillo:
Findings from the review of #206 that were triaged as real but not load-bearing for the release. Recorded here so they are not lost.

Each was verified against current code before being deferred; none of them affects released behaviour.

## Test isolation and assertions

- `cmd/gofastr/cov_migrate_test.go` — two new tests read the ambient `DATABASE_URL` instead of setting their own, so a developer with one exported runs them against a different database than CI does.
- `cmd/gofastr/examples_blueprint_test.go` — the fallback response is parsed without checking status first, so a non-200 produces a decode error rather than naming the real failure.
- `cmd/gofastr/lint_module_collision_test.go` — the lint is asserted against one blueprint. Every shipped blueprint declares an `app.module` under the framework's own module path, so all of them are fixtures for this rule.

## Tooling correctness

- `cmd/repolint/main.go` — the comment scanner treats `/*` inside a line comment or a string literal as the start of a block comment, so it can stop scanning a file early and silently under-report.
- `cmd/mutate/main.go` — `packageFiles` assumes `go list` returns one package. A pattern like `./framework/...` produces several `Dir` lines, and the parser reads the extra ones as file names, failing later with a confusing `read <dir>/<abs-path>` instead of naming the cause. Reject a multi-package pattern explicitly.

## Not actionable as filed

- `examples/ecommerce/app/axe_test.go` — a suggestion to change generated output. The file is emitted by `gofastr generate` and byte-compared by `TestCommittedAppMatchesGeneratorOutput`, so a hand edit is overwritten and fails the gate. If the change is wanted it belongs in `renderBlueprintAxeTest`.

## Rejected

One finding claimed the new `-race` gate would catch an unsynchronized `bytes.Buffer` read in `cmd/gofastr`. That package is not in the gate's package list and `-race` did not report it. The underlying race was real and was fixed in #206 anyway, since the access is unsynchronized whether or not a detector observes it.

Issue #207, opened 2026-08-19
No repair path for a database carrying a pre-v0.67 owner-column foreign key


DonaldMurillo:
Surfaced by the `historical upgrade fixtures` gate while landing #206.

## What happens

An entity that declares the same column as both `Scope.OwnerField` and a relation gets a foreign key from pre-v0.67 `AutoMigrate`:

```yaml
- name: tasks
  owner_field: user_id
  fields:
    - {name: user_id, type: relation, to: users}
```

That emits `FOREIGN KEY (user_id) REFERENCES users(id)`. The framework then stamps `user_id` from the session identity, which `battery/auth` stores in `auth_users` — so the column holds ids that are not keys into `users`. The constraint is one the framework violates on every create.

SQLite never checked, so it went unnoticed. Postgres has always rejected it.

## Why it needs a repair path

v0.67.0 does two things: it enforces foreign keys on every SQLite DSN, and it stops emitting this constraint. The second does not help a database that already has it. SQLite has no `DROP CONSTRAINT`, and `framework/migrate/schema_diff.go` explicitly has no table-rebuild path — it flags the need in a comment and stops.

So an affected app upgrading to v0.67.0 sees every create fail. The only remedy today is `_pragma=foreign_keys(0)`, which disables enforcement for the whole database. `evals/upgrade-fixtures` boots that way and proves the path works, but disabling the feature is a workaround, not a fix.

## What would close it

A SQLite table rebuild that drops the constraint: create the replacement, copy the rows, drop the original, rename. Either as a migration `AutoMigrate` can perform, or as a `gofastr` command that reports affected tables and offers to rewrite them.

Detecting affected tables is cheap — the schema is in `sqlite_master`, and the entity declarations say which column is the owner field.

## Evidence

`evals/upgrade-fixtures/fixtures/v0.53.0-app/seed.db` carries the shape. `fixtures/manifest.yml` documents that `users` is empty by design because auth lives in `auth_users`.

PR #212, opened 2026-08-20
Go 1.27 adoption: toolchain, stdlib uuid, CutLast, goroutine-leak profile, synctest

DonaldMurillo:
Rebased onto v0.69.0. This is now **Go 1.27 adoption only**.

## What changed in the rebase

The original 24 commits split cleanly: the first 12 were already ancestors of main, having shipped in **v0.68.0 via #221**, so `git rebase --onto main 9f84e9f0` dropped them and replayed the Go 1.27 dozen. Issues #207–#211 were closed by that merge on 2026-08-22 and are no longer claimed here.

Two conflicts:

- **CHANGELOG.** The original entry was written when `[Unreleased]` still held the v0.68.0 work, so most of it (ReadScope, `migrate repair`, the race-detector split, five fixes) is text already published under `## [0.68.0]`. Only the two genuinely new entries were kept under `[Unreleased]`: **Go 1.27**, and the **`go fix` `embedlit` known issue**.
- **`coreCongestionWindowGZ`.** v0.69.0 moved this line to 14464 for a 5-byte correctness fix in the anchor-click guard. This PR's re-baseline to 14700 supersedes it, and its reasoning is the better one: nothing in the framework compresses `runtime.js`, so Go's BestSpeed is a worst-case proxy for the nginx/CDN zlib that actually serves it, and Go 1.27 made the proxy pessimistic rather than making the artifact bigger.

## Two corrections to the re-baseline commit

Both from measuring, not reading:

**The numbers were stale.** They described a 41812-byte artifact compressing to 14602. v0.69.0's click-path work has since taken it to **41968 raw / 14660 compressed under Go 1.27.0**, so 14700 clears the real bundle by **40 bytes, not 98**.

**The ceiling is now measured rather than asserted.** `TestCoreBudgetRejectsCliffOverflow` brackets this constant from above; moving it and watching that test puts the boundary between **14800 (still guards)** and **14825 (vacuous)**. At 40 bytes of margin, the next feature needs a carve rather than another raise — which is what the policy on `TestTypicalPagePayloadBudget` (added in #223) already says.

## Verification

- Full suite green under Go 1.27.0 — 197 packages.
- `TestExampleBlueprintsBoot` failed in my first run and is **not** a regression: the generated apps booted fine but on a different port than the test polled, because `framework/isolation` remaps `PORT` when it detects a linked git worktree, and I ran the suite in one. Proven three ways — the Go 1.27 wave touches zero isolation files, the gate passes on a normal checkout, and it passes on this branch under Go 1.27 with `GOFASTR_ISOLATION=off`.
- The pre-push gate ran and passed.

This will be the repo's first CI run on Go 1.27, so a red check here deserves reading rather than a rerun.


DonaldMurillo:
Rescoped after #221 merged as v0.68.0.

The #207-#211 and #204 work shipped in #221 — this branch was split at the merge commit so CodeRabbit could review the hand-written half (its 626 files were past the 100-file review limit, 532 of them `go fix` output). What remains here is the Go 1.27 adoption on its own: toolchain move plus the modernizers, stdlib `uuid`, `strings.CutLast`, the goroutine-leak profile on the debug and MCP planes, synctest bubbles for the in-memory timing tests, and three baselines Go 1.27 moved (the level-1 gzip line, the TUI coverage floor, govulncheck to v1.7.0 so it can scan 1.27 source).

Retitled to match. Needs a rebase on main now that v0.68.0 has landed.

PR #221, opened 2026-08-21
Close #207-#211 and land #204

### LINKED ISSUE #207: No repair path for a database carrying a pre-v0.67 owner-column foreign key

### LINKED ISSUE #208: The framework root package has no race coverage, and it holds the goroutine-heaviest code
#206 added a `-race` gate to CI. The `framework` root package is excluded from it, and the reason is written at the gate in `.github/workflows/ci.yml`.

## The gap

`framework` holds `app.go`, `health.go`, and the process-module supervisor — the code in this repo that spawns and coordinates the most goroutines. None of it has race coverage.

The gate covers `battery/queue`, `battery/cache`, `framework/outbox`, `framework/fanout`, `core/fanout`, `framework/cron`, `framework/event`, `framework/hook`, `framework/ratelimit`, `framework/lifecycle`, `framework/slowquery`, `core/stream`, `framework/uihost`. The one package most likely to carry a data race is not on the list.

## Why it is excluded

The package also holds chromedp browser e2e tests and supervisor timing tests whose deadlines do not absorb race instrumentation. `TestSupervisor_*` has been seen flaking under `-race` while passing repeatedly in isolation, and a full-package `-race` run timed out `TestUIE2E_OwnerScope_CrossUserIsolation`.

That is slowdown, not a data race, but a gate that goes red for timing teaches people to ignore it.

The flake is real and load-dependent: while landing #206 the pre-push hook rejected a push on `TestUIE2E_OwnerScope_CrossUserIsolation` timing out at 61s, and the same test passed in 3s in isolation on the same machine minutes later.

## What would close it

Split the browser and timing suites into their own package so the remainder of `framework` can run under `-race`. The concurrency code and the tests that cannot tolerate instrumentation do not need to live together.

## Why it matters

The recurring defect class in this repo's release history is concurrency: queue claim/lease fencing, per-consumer outbox cursors, fanout subscriber queues. Every one of those was found by an audit reading code, never by a test.

### LINKED ISSUE #209: A read posture is all-rows or no-rows: no way to say "anonymous sees published rows only"
Surfaced three times in one review pass on #206, on three separate example blueprints.

## The gap

`Exposure.Access.Read` decides whether a caller may read an entity. It cannot say which rows. So an entity is readable in full or not at all, and there is no way to express the most ordinary content posture there is: anonymous visitors see published rows, signed-in editors see drafts.

## Where it bites

Three shipped examples open anonymous reads on entities whose own schema marks some rows unpublished:

- `examples/portfolio` — `posts.status` defaults to `draft`; `testimonials.approved` defaults to `false`
- `examples/project-manager` — `projects.is_public` defaults to `false`
- `examples/lms` — `courses` carries `draft` and `archived`; `modules` and `lessons` carry `draft`

Each serves those rows to anonymous callers through both the API and any screen. The blueprints now carry a `KNOWN EXPOSURE` comment saying so, because documenting it at the copy site was the only honest option available. People copy example blueprints as a starting point.

The alternative was gating reads entirely, which turns the public marketing pages of all three examples into an access notice.

## What would close it

A row-level read predicate that a posture can carry, applied by the same layer that applies owner and tenant scoping. `framework/crud`'s scope filters already compose predicates per entity for owner, tenant, and soft delete; a declared read filter would follow the same path.

Once it exists, the three examples should use it and the `KNOWN EXPOSURE` comments should go.

## Related

A narrower version of the same shape: a blueprint cannot express "anonymous read, any signed-in user writes". A blank `read:` leaves `AccessControl` empty so `Declared()` is false and the session requirement applies to reads too, and naming a permission does not help because a generated app grants permissions only to the admin role. `examples/ecommerce` and `examples/lms` both name that gap in comments.

### LINKED ISSUE #210: Nested filters refuse an owner filtering their own rows
Raised in review of #206 and deliberately not fixed there, because it restores a capability rather than closing a hole.

## Current behaviour

`checkNestedFiltersReadable` refuses `?rel.field=` outright when the target entity is owner-scoped or multi-tenant, unless the caller already holds cross-owner or cross-tenant access for that axis.

The reason is real: `buildExistsSubquery` emits no owner or tenant predicate, so the `EXISTS` clause counts every row in the target table. It does not return them, but the parent's row count changes with the guessed value, which is an oracle over any column. A signed-in owner could count every other owner's rows one guess at a time.

## The cost

An owner filtering *their own* rows through a relation is refused too, and that is a legitimate query. The workaround is the target entity's own list route, which is fine but means `?rel.field=` silently does not work for the most common scoped case.

## What would close it

Emit the owner and tenant predicates into the subquery, the way the include path already does per node in `applyRelatedOwnerScope` / `applyRelatedTenantScope`. A subquery narrowed to the caller's own rows is not an oracle, so the refusal could be dropped for the narrowed case.

The predicate value comes from the same context the include path reads. The care needed is on the anonymous and missing-owner paths, where the include path intersects rather than opting out, so a forged value matches nothing.

## Note

The exemption in this function is per-axis as of #206. A cross-owner grant does not lift a tenant refusal and vice versa; `TestCheckNestedFiltersReadable_ScopedTargets` covers both mixed cases. Any change here must keep that property.

### LINKED ISSUE #211: Deferred review findings from PR #206

DonaldMurillo:
Closes #207
Closes #208
Closes #209
Closes #210
Closes #211

Also picks up #204, which was opened 2026-08-17 and never landed.

Split out of #212 so it is reviewable: that PR carries the Go 1.27 adoption
too, and its 626 files put it past CodeRabbit's 100-file limit, so the review
skipped. 532 of those files were `go fix` output. This is the hand-written
half.

## #210 — a nested filter refused an owner filtering their own rows

`?rel.field=` compiles to an `EXISTS` clause that counts rows without selecting
them, so it could not narrow them by selecting, and the previous fix refused
the shape outright for every owner-scoped or multi-tenant target. That closed
the count oracle and took the ordinary case with it.

The subquery now carries the caller's owner and tenant predicates, built by the
same code the include and eager paths use, so the three surfaces cannot drift
into three different answers. Proven against a live database in both
directions: alice can filter by her own comment, and a guess at bob's comment
reads identically to a guess at a value that exists nowhere. Two mutations,
dropping the predicates and emitting them after the field predicate, each turn
it red.

Worth a reviewer's eye: `QueryBuilder` renumbers placeholders positionally by
encounter, so the scope values must be appended in the order their placeholders
appear, not in struct order. Getting it backwards binds the owner id to the
field predicate and returns nothing, which reads as "no matching rows".

## #207 — no repair path for a pre-v0.67 owner-column foreign key

Releases before v0.67 emitted a key on a column the framework stamps from the
session identity, which lives in `auth_users`, so it referenced a table where
no matching row will ever exist and every create violated it. SQLite never
checked; Postgres always rejected it.

`gofastr migrate repair` reports by default and rewrites under `--apply`. The
replacement DDL is the original minus the offending clause rather than a fresh
`CREATE TABLE` from the entity declaration, because a legacy table carries
columns, defaults and constraints the declaration no longer mentions and a
repair that loses data is worse than the constraint it removes. Indices and
triggers are replayed. Other foreign keys keep enforcing and a composite key
that merely includes the owner column is left alone.

Enforcement is off for the rebuild by necessity, on a pooled connection that
goes on to serve ordinary writes. If it cannot be turned back on the connection
is destroyed rather than returned.

## #208 — the framework root package had no race coverage

It holds `app.go`, `health.go` and the process-module supervisor. The exclusion
was never about a data race: two suites whose deadlines do not absorb race
instrumentation lived there, and they moved to `framework/uie2e` and
`framework/processmoduletest`. The root package runs 22s under `-race`, down
from 48s. Proven by injecting an unsynchronised counter and watching the
detector report it.

The `./framework` match:/processmodule floor moves 84.0 to 80.3: the moved
suites' coverage is no longer attributed there. Measured 82.5 after the move.

## #209 — a read posture was all-rows or no-rows

`Exposure.ReadScope` narrows which rows a caller may read. One builder feeds
every read: list data and count, get, cursor, stream, the in-process API, typed
queries, the upsert read-back, and the include and eager loaders at every
depth. A row outside the scope answers 404, not 403. Declared in a blueprint as
`read_scope:`, surviving the pack round-trip, with every level of the structure
going through the unknown-key check.

The three examples that carried a `KNOWN EXPOSURE` comment now declare the
posture and the comments are gone.

Deliberately not covered, and stated in the code and the docs rather than left
implied: writes are not filtered, so a caller with write-but-not-read can
update or delete a row they cannot see, and `RETURNING` on an upsert is not
filterable without changing write semantics.

## #211 — deferred findings from PR #206

`repolint` stopped reading a file at a `/*` inside a line comment or string
literal, reporting clean for a file it had given up on. `make mutate`
mis-parsed a multi-package pattern. Three gates measured the developer's own
exported `DATABASE_URL`.

## Two bugs found while building this, both caught by something other than the suite

**A list filter narrowed the total but not the rows.** `filter.ApplyToQuery`
reached the count query and not the data query, so `?field=value` returned the
whole table under a correct-looking total. The entire crud suite stayed green,
because nothing asserted that a filter narrows the ROWS. An admin battery test
one package over is what caught it.
`TestListFilterNarrowsRowsAndTotalTogether` now pins the two queries against
each other and fails if either loses the filter.

**Three `crud: true` flags went missing** from lms `courses`, lms `modules` and
project-manager `boards` while `read_scope` was being added. Nil defaults to
true so nothing failed. Every entity's flag now matches the merge base exactly.

## The merge with main

main landed #213, the unslop sweep, mid-flight: 2176 files of comment rewrites
against this branch's 53. 2176 auto-merged, 18 files conflicted over 39 hunks.
Three things the resolution caught that a mechanical "take ours" would have
buried: a duplicated `SearchFields` reader in `pack.go`, the `crud: true` flags
above, and 36 em dashes that would have been silently reinstated inside the
test files this branch moved.

## Verification

`scripts/test-all.sh` green. Pre-push gate green: full suite, coverage floors,
module integrity. Every guard added here was proven by mutation, including the
ones the workers reported: each claim was re-run independently rather than
taken on trust.

## Summary by CodeRabbit

* **New Features**
  * Added configurable row-level read scoping across lists, counts, relations, includes, streaming, and typed queries.
  * Added blueprint and generated-configuration support for read-scope rules.
  * Added `gofastr migrate repair` to detect and repair stale SQLite owner foreign keys while preserving data and schema.
* **Bug Fixes**
  * Improved filtering, nested-scope authorization, package discovery, repository lint scanning, and environment precedence handling.
* **Documentation**
  * Expanded guidance for read scoping, migrations, security, and relation filtering.
* **Tests**
  * Expanded security, migration, and race-detector coverage.

## BREAKING for in-process callers (added after review)

`ListAll` and `CountAll` now apply owner, tenant, and read-scope predicates to
a `ListOptions.NestedFilters` spec on **every** call, not only over HTTP.

The narrowing used to be gated on a context marker no host could set:
`withRealRequest` is unexported, and crud's three call sites hand the marked
context to the include loader, which never reaches `ListAll`. The safe branch
was unreachable from every production shape, so the default was the unsafe one,
and the test that proved it worked only passed because it lives in-package.

A host handler forwarding a caller-influenced relation into `ListAll`
mid-request therefore kept the count oracle the HTTP path closes.

Server-side code that means "across every owner" now says so with
`owner.AllowCrossOwner` or `tenant.AllowCrossTenant` — both already exist for
this surface, both are unreachable from an HTTP route.

The posture check (may this caller read the target at all) stays HTTP-only,
which is where `ApplyIncludes` already draws the line: it is the baseline
session gate, in-process code has no session by construction, and running it
there refuses every background nested filter while protecting nothing.


DonaldMurillo:
@coderabbitai review

DonaldMurillo:
@coderabbitai review

DonaldMurillo:
Triage on the `assertColumnType` / `db.QueryContext` finding: **rejected**, with reasoning.

There is no `.golangci.yml` in this repo and no `noctx` check in any gate (pre-commit, pre-push, or CI), so nothing enforces context propagation here. `framework/migrate/sqlite_ownerfk_test.go` has 41 non-context `db.Query`/`db.QueryRow`/`db.Exec` calls and zero context ones; across the package's tests it is 27 to 0. The helper matches the file's convention exactly.

Converting one call site out of 42 would make the file less consistent, not more, and buys nothing concrete: the value would be cancellation on test timeout for a single-row pragma query that takes microseconds.

Happy to convert the whole file if context propagation in tests is wanted as a rule — that's a coherent change. One site is drift.

Issue #220, opened 2026-08-21
GOFASTR1801 does not say it is flagging the identifier, not the value


danansa86:
**What happened, and what did you expect?**

I named a Go variable `fill` — it held a colour for an SVG attribute. The build
failed with GOFASTR1801.

I read the message as being about the value, so I went looking at what I was
assigning. The rule is about the identifier: it flags Go identifiers named
after CSS properties.

Once that was clear the fix took ten seconds — `fill` became `face`, `stroke`
became `edge`. Working out which half the rule meant took considerably longer.

**Versions**

- gofastr: v0.67.0
- go: go1.26.5 darwin/arm64
- OS / DB: macOS + SQLite

**Minimal repro**

```go
face := "var(--color-surface)"  // fine
fill := "var(--color-surface)"  // GOFASTR1801
```

**Suggestion**

Name the identifier in the message, e.g.

> GOFASTR1801: identifier `fill` is named after a CSS property; rename it

The rule looks right. Only the message needs to say which half it is about.

Issue #219, opened 2026-08-21
gofastr dev serves stale project JS: the file it returns and the file it runs disagree


danansa86:
**What happened, and what did you expect?**

I edited a project `.js` file and hard-reloaded. The old code was still running.
This happened three times in one session, and each time I spent a while
assuming my change was wrong before checking whether it had loaded at all.

What settled it: `fetch('/nav.js', {cache:'reload'})` returned the new file
while the executing code was the old one. The server had the change and the
page did not. Only restarting `gofastr dev` cleared it.

Framework assets are versioned — `/__gofastr/runtime.js?v=00fac3ae373ff996`.
Project assets are served bare: `/nav.js`, `/app.css`.

**Versions**

- gofastr: v0.67.0
- go: go1.26.5 darwin/arm64
- OS / DB: macOS + SQLite

**Minimal repro**

1. `gofastr dev`
2. Edit a project JS file — a `console.log` is enough.
3. Hard reload.
4. Old code.

**Suggestion**

Version project static assets in dev the way framework assets already are, or
send `Cache-Control: no-store` for them while `dev` is running.

The cost is out of proportion to the fix: you do not doubt the cache, you doubt
your own change and go looking in the wrong file.

Issue #218, opened 2026-08-21
activelink strips aria-current and stamps "page", and neither is documented


danansa86:
**What happened, and what did you expect?**

Two hours lost to the same runtime, in opposite directions.

First: I set `aria-current` on nav links for a current-section indicator. It
kept getting cleared. `activelink` walks every `nav a` and removes
`aria-current` from anything whose href is not the current path — which an
in-page anchor never is.

Second, later: I wrote a CSS rule for `[aria-current="true"]` and it never
matched. `activelink` stamps `aria-current="page"`, which is the correct ARIA
value. My rule was wrong, and a selector that matches nothing is not an error.

Neither is a bug. Both cost real time because the contract is not written down.

**Versions**

- gofastr: v0.67.0
- go: go1.26.5 darwin/arm64
- OS / DB: macOS + SQLite

**Minimal repro**

1. Set `aria-current` on a nav link whose href is not the current path.
   Navigate. It is gone.
2. Write `.nav-link[aria-current="true"] { … }` and load a page the nav links
   to. It does not apply — the value is `page`.

**Suggestion**

Three sentences in the docs: what `activelink` sets, which value it uses, and
what it removes. That would have covered both hours.

Issue #217, opened 2026-08-21
The SPA router preventDefaults in-page anchors before a userland handler sees them


danansa86:
**What happened, and what did you expect?**

I added a click handler for in-page anchors — smooth scroll to a section.
Clicking a nav item changed the URL and the page did not move.

My handler checked `e.defaultPrevented` and stood down, which is the polite
thing for a userland handler to do. The SPA router had already called
`preventDefault` on the way past: it treats `/#section` as a navigation to `/`,
prevents the default and updates the address bar.

So the link looked broken while every part of it was working.

**Versions**

- gofastr: v0.67.0
- go: go1.26.5 darwin/arm64
- OS / DB: macOS + SQLite

**Minimal repro**

1. Add a bubble-phase `click` listener on `document`.
2. Return early if `e.defaultPrevented`.
3. Click an in-page anchor such as `/#features`.

The fix on my side was capture phase plus `stopPropagation`, which means racing
the router rather than cooperating with it.

**Suggestion**

A documented, cancellable `gofastr:beforenavigate` event, so userland can claim
a click without having to get in front of the router.

PR #222, opened 2026-08-23
v0.69.0: seven v0.67.0 field reports on silent failures

### LINKED ISSUE #214: A theme token misspelled in project CSS is inert, and nothing anywhere says so
**What happened, and what did you expect?**

Every rounded corner on the site rendered square. No error — not in the build,
not in the console, not from the linter.

The cause was a name. The theme emits `--radii-lg`. I had written `--radius-lg`.

An invalid `var()` is not an error in CSS. It resolves to nothing and the
declaration is dropped. Fifteen declarations were inert for days.

I expected the typed theme to reach this. It checks the definition — a
`style.Radius` is compiler-checked — but the reference in a project stylesheet
is just a string, and nothing connects the two.

**Versions**

- gofastr: v0.67.0
- go: go1.26.5 darwin/arm64
- OS / DB: macOS + SQLite

**Minimal repro**

1. Define a `RadiusSet` in the theme.
2. In a project CSS file: `border-radius: var(--radius-lg);`
3. Build and load. No error, no radius.

**Suggestion**

GOFASTR1801 exists, so the hook is already there. Either emit a token manifest
the build can lint project CSS against, or have the linter flag any `var(--…)`
in project CSS that the theme does not emit.

DonaldMurillo:
Closes #214, #215, #216, #217, #218, #219, #220.

Seven reports from one session of building against v0.67.0. Most of them
cost hours because the failure was silent: an invalid `var()`, an edit to
the palette nobody reads, a cache nobody suspects, a linter naming the
wrong half of the line. One commit each.

## What changed

**#219 — `gofastr dev` served stale project JS/CSS.** Framework assets are
content-versioned; project assets serve at plain URLs with a bare
`Last-Modified`, which invites heuristic caching. The browser kept running
the old script while a fresh fetch returned the new bytes. Under
`gofastr dev` they now go out `no-store`, from both the static dir and the
embedded FS. Production is untouched.

**#215 — editing `Colors` did nothing when `DarkColors` is what renders.**
`style.DarkPaletteGaps` returns the tokens a non-empty `DarkColors` omits
and the host logs them once at boot. The framework theme now declares dark
values for the three code-surface tokens it had been leaving out on
purpose, so the default stays silent and the intent is written down rather
than implied by an omission.

**#216 — `app.Layout`'s `<header>` wrapper stopped a sticky header.** A
sticky element only travels inside its parent's box, and that parent was
65px tall. `WithStickyHeader()` makes the wrapper the sticky element, on
the theme's `--z-sticky` layer with a background so content does not scroll
through it. The banner landmark stays.

**#217 — the router `preventDefault`ed before userland saw the click.**
New cancellable `gofastr:beforenavigate`, fired on the anchor after every
fall-through check. Cancelling it means the router stands down completely
while still suppressing the browser default, so a claimed click never
becomes a hard page load.

**#218 — `activelink` cleared an `aria-current` another module owned.**
Links inside a `[data-fui-scrollspy]` wrapper are now left alone, and
`data-fui-activelink-skip` is the general opt-out. Both attribute tables
now state what the module writes, what value it uses, and what it removes.

**#220 — GOFASTR1801 read `fill := "var(--color-surface)"` as CSS.** The
`:` of `:=` satisfied the property colon. Matches containing `:=` are
rejected, and the message names the fragment that matched.

**#214 — a misspelled theme token was inert and nothing said so.** New
`GOFASTR1806`. `contracts.Pass` discovers `.css` files and serves them
through `StyleFiles()`; the Go accessors stay Go-only. The rule reports a
`var(--name)` the theme does not emit and suggests the nearest token.

## Three things worth a reviewer's attention

**#215 is deliberately a partial fix.** The reporter's exact case, editing
the light value and forgetting the dark one, is not detectable at runtime:
nothing records what a token used to be, so there is no baseline. What
ships is the detectable half (an incomplete dark palette) plus docs. I did
not merge the two palettes into one `Palette{Light, Dark}` type, which is
what would actually close the class; that is a breaking redesign and wants
its own decision.

**#220 turned out to be a false positive, not a wording problem.** The
issue asked for a clearer message on the assumption the rule was right.
It was not: the line is Go, and reporting it was a bug. Both halves are
fixed, and I would rather the issue be closed on the larger finding.

**#214 needed a suppression fix to be usable.** `collectSuppressions`
walked Go files only, so the one rule that fires on stylesheets had no
escape hatch in the file it fires in, and the non-Go comment scanner did
not recognise `/*`. Both fixed, with tests.

Two smaller judgement calls: `data-fui-scrollspy-target`'s rows in both
attribute tables were wrong (it is a selector on the wrapper, not a marker
on a link) and are corrected here; and a dead `anchor.target !== ''` clause
came out of the nav click handler to keep the level-1 runtime budget under
its cap rather than raising the cap.

## Verification

- Every guard proven by mutation, not by argument: inverted the dev
  predicate and watched the production tests fail; deleted the `:=`
  rejection, the `TokenNames` membership check, the fallback check, the
  beforenavigate bail-out, the activelink skip, the sticky modifier, and
  the CSS suppression fix, and watched exactly the intended tests fail
  each time.
- `./scripts/test-all.sh` green with `TEST_POSTGRES_DSN` set, including
  coverage floors and `gofastr verify --no-vet --strict`.
- The `core-ui/runtime` chromedp suite run isolated: 218s, green.
- #216 checked in a browser, not only in assertions: after scrolling
  1200px the header's `getBoundingClientRect().top` is 0 with the modifier
  and -1200 without, and the background occludes the content beneath it.
- Every commit builds, vets, and passes its gates on its own, checked in a
  detached worktree, because several were staged by hunk and the pre-commit
  hook only ever sees the full worktree.

## Summary by CodeRabbit

- **New Features**
  - Added optional sticky layout headers with themed positioning, layering, and backgrounds.
  - Added CSS theme-token validation with typo suggestions and stylesheet scanning.
  - Added dark-palette gap detection and startup warnings for incomplete palettes.
  - Added cancellable pre-navigation events, active-link opt-outs, and improved scrollspy support.

- **Bug Fixes**
  - Prevented development assets from being cached.
  - Reduced false-positive CSS findings.
  - Improved case-insensitive CSS keyword matching while preserving custom-property name sensitivity.
  - Improved handling of unsupported and SVG link targets.
  - Prevented scrollspy and active-link state interference.

- **Documentation**
  - Expanded guidance for layouts, navigation, theming, live reload, CSS scanning, and the 51-rule catalog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
 
 ## [Unreleased]
 
+## [0.69.0] - 2026-08-23
+
 ### Added
 
 - **`gofastr verify` catches a theme token that does not exist
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,10 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   declaration only inside a rule block and outside parentheses, so the
   condition of a feature query (`@supports (--brand: red)`, which asks whether
   the browser can parse that declaration rather than making one) no longer
-  marks the token known; `@property --name` registrations do count. `style.TokenNames()` is
+  marks the token known; `@property --name` registrations do count. The
+  `var(` and `@property` keywords match case-insensitively, as CSS does, while
+  the token name stays case-sensitive, as CSS also does: `VAR(--radii-lg)`
+  resolves, and `var(--mixed)` against a declared `--Mixed` does not. `style.TokenNames()` is
   the manifest it checks against, built from the same reflection walk that
   emits the CSS, so the two cannot drift apart. The scan is string-aware:
   `content: "var(--x)"` is content rather than a reference, and a quoted `/*`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,11 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   when one is within an edit distance of two. It stays quiet for a reference
   carrying a fallback, because `var(--x, 8px)` degrades instead of dropping
   the declaration; for a custom property declared in any scanned stylesheet;
-  and for the `--ui-*` and `--fui-*` component knobs. `style.TokenNames()` is
+  and for the `--ui-*` and `--fui-*` component knobs. A `--name:` counts as a
+  declaration only inside a rule block and outside parentheses, so the
+  condition of a feature query (`@supports (--brand: red)`, which asks whether
+  the browser can parse that declaration rather than making one) no longer
+  marks the token known; `@property --name` registrations do count. `style.TokenNames()` is
   the manifest it checks against, built from the same reflection walk that
   emits the CSS, so the two cannot drift apart. The scan is string-aware:
   `content: "var(--x)"` is content rather than a reference, and a quoted `/*`
@@ -67,6 +71,15 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   becomes a hard page load, and then does nothing else. No `pushState`, no
   partial fetch, no intercept overlay.
 
+  The same guard now matches the `target` keywords the way HTML does,
+  ASCII-case-insensitively: `target="_SELF"` names the current browsing
+  context exactly as `_self` does, and comparing the attribute raw sent it
+  down the skip path, turning a soft navigation into a full page load. The
+  core runtime's level-1 budget moved 128 bytes (14336 to 14464) to carry the
+  five bytes that costs, and the reason sits beside the constant: the usual
+  answer of carving a feature into a demand module is the one thing
+  unavailable for a guard that IS the click path.
+
 - **`data-fui-activelink-skip` opts a nav link out of active-route
   highlighting.** The `activelink` module neither sets nor clears
   `aria-current` or `.active` on a link carrying it, at load or after
@@ -84,11 +97,14 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   both compile, and nothing said which one was being read.
   `style.DarkPaletteGaps` returns the color tokens a non-empty `DarkColors`
   has no usable dark value for, missing keys and keys present with an empty
-  value alike (an empty one emits `--color-surface: ;`, which the browser
-  drops, so it fails exactly as a missing key does), and the UI host logs them
-  once at boot: those tokens keep their light
-  values in dark mode, which is usually a contrast bug. An empty `DarkColors`
-  is a light-only theme by design and stays silent. The framework theme now
+  value alike, and the UI host logs them once at boot. A missing key leaves
+  the light value painting in dark mode, which is usually a contrast bug. An
+  empty value is worse: an empty custom property is valid CSS, so
+  `--color-surface: ;` overrides the light declaration and every
+  `var(--color-surface)` substitutes to nothing, dropping the consuming
+  declaration at computed-value time. The token paints transparent rather
+  than light. An empty `DarkColors` is a light-only theme by design and stays
+  silent. The framework theme now
   declares dark values for `code-surface`, `code-text`, and `code-border`
   matching its light ones, which was always the intent and existed only as an
   omission.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,10 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   the declaration; for a custom property declared in any scanned stylesheet;
   and for the `--ui-*` and `--fui-*` component knobs. `style.TokenNames()` is
   the manifest it checks against, built from the same reflection walk that
-  emits the CSS, so the two cannot drift apart. A stylesheet can waive it with
+  emits the CSS, so the two cannot drift apart. The scan is string-aware:
+  `content: "var(--x)"` is content rather than a reference, and a quoted `/*`
+  does not open a comment that swallows the rest of the file along with the
+  real typos in it. A stylesheet can waive it with
   `/* gofastr:allow(GOFASTR1806) reason */`, which meant teaching the
   suppression scanner to read `.css` files at all: it walked Go source only,
   leaving the one rule that fires on stylesheets with no escape hatch in the
@@ -43,10 +46,14 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   `layout--sticky-header` modifier and `LayoutBaseCSS` sticks the wrapper on
   the theme's `--z-sticky` layer, with a background from
   `--ui-layout-header-bg` so page content does not scroll visibly through it.
-  The wrapper and its banner role stay. `layouts.md` now writes down what each
-  layout slot is wrapped in, which is the fact whose absence cost the time.
+  The wrapper and its banner role stay. The sticky shell is also a flex column
+  that fills the viewport rather than exceeding it: `.layout-body` carries
+  `min-height: 100vh`, so a header above it made a short page scroll by exactly
+  the header's height with nothing in the overflow, which under a pinned header
+  reads as broken. `layouts.md` now writes down what each layout slot is wrapped
+  in, which is the fact whose absence cost the time.
 
-- **`gofastr:beforenavigate`, a cancellable event the SPA router fires before
+- **`gofastr:beforenavigate`, a cancelable event the SPA router fires before
   it takes a click.** A userland click handler that checked
   `e.defaultPrevented` and stood down, which is the polite thing for a
   userland handler to do, never ran: the router had already called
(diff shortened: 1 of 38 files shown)


DonaldMurillo:
Triaged all seven. Six fixed in 4d172f22, one declined with measurements. Each was verified against the code before being acted on.

**Fixed**

- **`rendering.go` — CSS strings.** Confirmed both directions. `content: "var(--not-a-token)"` reported on valid CSS at `SeverityError`, i.e. failed a build; and a quoted `/*` opened a comment that swallowed the rest of the file, hiding the real `--radius-lg` typo after it. The pre-pass is now a four-state scanner (text / comment / `'…'` / `"…"`). It blanks bodies to spaces and leaves newlines in place, so byte offsets and reported line numbers survive; `cssblank_test.go` asserts that invariant directly across unterminated strings, escaped newlines and a trailing backslash at EOF, since a span that changes length silently misattributes every later finding. I did not reach for a tokenizer — this is a pre-pass for two regexes.

- **`darkpalette.go` — empty values.** Confirmed: `darkSchemeCSS` writes an empty value as `--color-surface: ;`, which the browser drops, so it fails exactly as a missing key does. Now a gap, with a regression test proven by mutation.

- **`layout_css.go` — sticky shell height.** Confirmed the overflow, but not the attribution: it measures 64px **with and without** `WithStickyHeader()`, so it is pre-existing in the plain-header shell, not introduced here. `.layout-body` carries `min-height: 100vh` and the contained and sidebar shells already each work around it; the plain one was overlooked. Fixed under the new opt-in modifier only, so no existing surface changes behavior. Measured at 900x800 with a 64px header: scrollHeight 864 → 800, long page still scrolls (3440) with the header at `rect.top` 0, non-sticky shells unchanged. **The general case is still there for any plain layout with a header** — happy to open a separate issue for it, since fixing it globally is a design-system-wide change that wants Meridian verification.

- **`activelink_skip_e2e_test.go` — retention.** Good catch: the protected links started without `class="active"`, so the test only proved the module does not *add* it, never that it does not *remove* an author-set one, which is the same branch that clears `aria-current`. Fixture now starts them with the class and asserts retention, plus a bare skip link that must gain nothing. Both new assertions fail independently when the skip check is removed.

- **`dev-livereload.md` — accuracy.** Agreed, the claim was over-broad. True for the static dir (read from disk per request), false for an embedded static FS (compiled into the binary). The doc now says which is which.

- **`CHANGELOG.md` — spelling.** Now `cancelable` throughout, matching the `CustomEvent` option it names.

**Declined, with measurements**

- **`nav.js` — case-insensitive `target`.** Real gap; `_SELF` should match `_self` per the HTML spec. But the core runtime is at **14334 bytes against a 14336-byte cap** enforced by `TestRuntimeModuleSizeBudgets`, whose failure message says to carve a feature into a demand module rather than raise the line. Both forms measure over: `toLowerCase()` → 14339, and the shorter `/^(?!_self$)./i` regex → 14341 (the shorter source compresses worse, since gzip already folds the repeated `anchor.target` tokens). The failure mode is a normal same-tab navigation instead of an SPA one, on an anchor spelling that does not occur in practice. Not worth 2 bytes of the initial congestion window; if it should ship, it needs bytes carved elsewhere first.

DonaldMurillo:
Second round: both findings fixed, and I've reversed the one I declined last round.

**The `target` guard — reversed, now fixed.** Last round I declined the case-insensitive `target` match on budget grounds. That was the wrong call, and the budget moved instead: 14336 → 14464 bytes at gzip level 1, with the reason written beside the constant rather than left in a commit message. The comment there tells the next person to carve a feature into a demand module instead of raising the line; that advice holds and stays, with the exception named — this guard *is* the anchor-click path, which the same comment already says must stay in core, so carving was never available for it. 14464 is still under the figure the cliff comes from (10 packets at a 1460-byte MSS, ≈14600 bytes).

Two tests, both proven by mutation:
- `target="_SELF"` is soft-navigated. The assertion is a `window` stamp surviving the click, which distinguishes "the router handled it" from "the browser did and happened to land on the same URL". Reverting to the raw comparison fails it.
- `target="_blank"` still escapes the router entirely, and `gofastr:beforenavigate` does not fire for it. Deleting the guard fails it. That is the assertion that stops a fix for the first case from quietly hijacking every target.

**`@supports` conditions — fixed.** Confirmed the false negative first: `@supports (--radius-lg: 0) { .a { border-radius: var(--radius-lg) } }` reported nothing, so the feature query silenced a real typo. A `--name:` now counts as a declaration only inside a rule block and outside parentheses. I added the opposite-direction test too, because over-narrowing here is the worse failure: a declaration nested inside `@media` *and* `@supports` blocks still counts, and `@property --name` registrations (no colon after the name, so the declaration scan could never see them) now count as well. All three fail under their own mutations.

**Empty custom properties — you were right, and my explanation was wrong in the direction that matters.** I claimed an empty value is a malformed declaration the browser drops, so it "fails exactly as a missing key does". Measured in Chrome against a green light value:

```
--tok: ;      (re-declared empty)  → background-color: rgba(0, 0, 0, 0)
              (not re-declared)    → background-color: rgb(0, 128, 0)
getPropertyValue('--tok') on the empty one → ""
```

So an empty custom property is valid CSS and *overrides* the light declaration; `var()` substitutes to nothing and the consuming declaration is dropped at computed-value time. Empty is the **worse** gap, not the equivalent one — a missing key keeps the light value, an empty one paints transparent. The code was already right to report both; the doc comment, the test comment, and the changelog entry were not, and all three now say what the browser actually does.

One thing I looked at and did not change, flagged rather than silently skipped: `anchor.target` reflects the attribute only, so with `<base target="_blank">` on the page an anchor carrying no `target` still takes the SPA path. Worth its own issue if it matters; it is a different guard from the one under review here.

DonaldMurillo:
@coderabbitai review

DonaldMurillo:
Self-review of the last commit, since CodeRabbit is incremental and rate-limited and never saw it — it found a regression I had introduced in that same commit.

`anchor.target.toLowerCase()` assumed an `HTMLAnchorElement`. An SVG `<a href>` matches the `a[href]` selector and `closest()` reaches it from a child (both verified in Chrome), and `SVGAElement.target` is an `SVGAnimatedString`, not a string — so `.toLowerCase()` throws, aborting the router's handling of that click and surfacing as an uncaught TypeError. The shape I replaced was accidentally safe there: a truthy object is never `'_self'`, so the click fell through to the browser.

`String()` first restores exactly that outcome for SVG anchors while keeping the `_SELF` fix for HTML ones. The router has never claimed SVG links and this does not start it doing so. Regression test asserts no uncaught error reaches `window.onerror` and that the link is not navigated in place; dropping the `String()` fails it with `anchor.target.toLowerCase is not a function`.

Worth noting for the record: the `target` finding was the one I declined on budget grounds last round, and taking it did cost two rounds and a 128-byte budget move. The fix is right, but the review value here came from the mutation tests, not from the original suggestion.

DonaldMurillo:
Third finding fixed, plus one I found reviewing my own last commit.

**`var(` casing — fixed.** Right, and the miss runs both ways. CSS function and at-rule names are ASCII case-insensitive, so `VAR(--radius-lg)` resolved in a browser while the rule stayed silent, and an uppercase `@PROPERTY --brand` registration was not counted, which reported reading `--brand` as a typo.

The half that must *not* fold is the token name, and there is a tempting over-fix that does. Measured in Chrome:

```
VAR(--tok)  → rgb(0, 128, 0)     function name folds
Var(--tok)  → rgb(0, 128, 0)
@PROPERTY --reg { … }            at-rule name folds, registration works
var(--mixed) vs declared --Mixed → rgba(0, 0, 0, 0)   names do NOT fold
```

So the keywords fold and the names don't. There's a test for the name half specifically, and it fails under the mutation that lowercases both sides — that over-fix would call a real miss a match and hide the exact failure the rule exists for. All three casing guards proven by mutation.

**Self-review finding: I broke SVG links in the previous commit.** `anchor.target.toLowerCase()` assumed an `HTMLAnchorElement`. An SVG `<a href>` matches `a[href]` and `closest()` reaches it from a child, and `SVGAElement.target` is an `SVGAnimatedString` — `.toLowerCase()` throws, aborting the router's handling of that click. Fixed with `String()` first, which restores the previous outcome for SVG anchors (a stringified object is never `'_self'`, so the click falls through to the browser) while keeping the `_SELF` fix for HTML ones. Regression test asserts nothing reaches `window.onerror`; dropping the `String()` fails it with `anchor.target.toLowerCase is not a function`.

**CI note:** `build · vet · test` went red once on `99f40c6f` with `TestAppStart_WiresRunSeeds` timing out under `-race`. That test is in package `framework`; the commit touched only `core-ui/runtime` (nav.js, its generated composition, one e2e test). The same job passed on the three previous pushes of this branch, and the test passes locally under `-race -count=8`. Treating it as a flake and re-running rather than as a signal — flagging it here rather than staying quiet about a red run.

DonaldMurillo:
@coderabbitai review

DonaldMurillo:
@coderabbitai review

DonaldMurillo:
Rolled for release in 097a1029: CHANGELOG stamped `0.69.0 - 2026-08-23`, SECURITY.md moved to `0.69.x` so the release gate accepts the tag, and the upgrade registry's `through` bumped with an entry beside it.

Minor rather than patch because this line ships only minors, and because one of the three registry notes is actionable on upgrade even with nothing marked BREAKING: **GOFASTR1806 is a new ERROR-severity rule, so it can fail a `gofastr verify` run that passed before.** That is the same disruption a breaking API change causes without an API changing, so the note leads with it and lists all four ways out (fix the spelling, add a fallback, declare the property, waive it in the stylesheet).

The other two notes: the dark-palette warning is new while the rendering is not, so an upgrader seeing a fresh WARN at boot should not have to work out whether something changed underneath them; and the GOFASTR1801 fix needs no action but detects `gofastr:allow(GOFASTR1801)` directives, since anyone who suppressed the false positive now holds a stale suppression that verify reports.

Verified by running `gofastr upgrade` from v0.68.0 to v0.69.0 against a scratch module and reading what it prints, rather than only parsing the YAML.

PR #231, opened 2026-08-26
fix(ui): pin Card footers to the card's bottom edge

DonaldMurillo:
A config-only `ui.Card` (Heading/Description/Footer, no body children) has no flex-stretch element inside — `.ui-card__body` carries `flex:1` but only renders when body children exist. In any stretch context (a `ui.Grid` row, an align-stretch flex parent) the footer floated mid-card with dead space under it, and sibling cards' footer rows misaligned.

Caught screenshotting the relayboard recipe's pricing grid in gofastr-plugins#13: the Team card's buy row hung 93px above the card edge.

**Fix**: `margin-top: auto` on `.ui-card__footer`. Footer-less cards are untouched by the rule, so no existing surface moves — Meridian uses no Card footers (grep-verified) and its home page screenshots unchanged.

**Proof (Hard rule 11)**: the new chromium-tagged test renders a real two-card Grid in Chrome and measures each footer's bottom edge against its card's. Watched failing with the margin removed (93.0px gap on the config-only card), restored to green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Card footers now remain pinned to the bottom edge of equally sized cards, including linked cards in Grid layouts.
  * Analytics request tracking now preserves credential headers when browser events arrive out of order.

* **Tests**
  * Added browser coverage verifying footer placement across multiple card shapes and content configurations.

* **Documentation**
  * Updated release and upgrade notes for version 0.71.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
 
 ## [Unreleased]
 
+## [0.71.2] - 2026-08-26
+
 ### Fixed
 
 - **framework/ui: Card footers pin to the card's bottom edge.** A
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,11 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   no flex-stretch element inside, so in a stretch context — a `ui.Grid`
   row, an align-stretch flex parent — the footer floated mid-card with
   dead space beneath it and sibling cards' footer rows misaligned. The
-  footer now carries `margin-top: auto`; a chromium-tagged layout test
-  measures the pinned edge in a real Grid.
+  footer now carries `margin-top: auto`, and the linked variant's
+  `.ui-card__inner` wrapper grows to fill the stretched anchor so the
+  pin reaches the card edge there too (CodeRabbit caught the linked
+  sibling). A chromium-tagged layout test measures all three shapes'
+  pinned edges in a real Grid.
 
 ## [0.71.1] - 2026-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
 
 ### Fixed
 
+- **framework: a Shutdown during startup stops Start.** `App.Shutdown`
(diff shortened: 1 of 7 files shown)


DonaldMurillo:
CodeRabbit triage (all three findings reviewed on merit):

1. **Linked-card footer (Minor) — accepted, fixed in 90eeceeb.** Verified against card.go: with `Href` the footer nests in `.ui-card__inner`, which had no flex-grow, so the pin stopped one wrapper short of the anchor's edge. `.ui-card__inner` now carries `flex: 1 1 auto`; the chromium test renders the linked shape as a third grid card and was watched failing at a 129px gap with the grow removed.

2. **Test-owned layout CSS (Major) — accepted in part.** Viewport width moved out of page CSS into browser setup (`chromedp.WindowSize`), which was the substantive half. The remaining `<!doctype html>` + `body{margin:0}` harness is the established chromium-test idiom in this package (see `image_placeholder_chromium_test.go`); Hard rule 7 governs shipped surfaces, not test fixtures, and inventing a shared rendering-surface package is out of proportion for a two-rule reset. Rejected that half.

3. **Rendered-image validation (Major) — rejected.** `getBoundingClientRect` measures rendered layout from the compositor — it is exactly the machine-checkable form of this regression, and it sees what a DOM dump cannot. A stored screenshot with no assertion is dead weight in CI, and a pixel-diff of footer position is strictly more brittle than the geometry it would approximate. The pixels half of Hard rule 9 was honored where it belongs: the bug was found and re-verified by reading actual screenshots of the relayboard grid, and the changelog claim now matches the fixed behavior including linked cards.

DonaldMurillo:
Round-2 triage:

1. **Same-row anti-vacuity guard — accepted, fixed.** Each card's `top` is now asserted equal to the first's before the gap checks; watched firing on a 300px viewport (card 1 wrapped, top 181 vs 0) and restored to green at 800px.

2. **Per-instance ExtraInfo queues for redirect chains — rejected.** The interleaving that misattributes headers requires a redirected request on an asserted path: the paths this suite asserts on are relay beacons, and the relay refuses upstream redirects by design (battery/relay), while the fake vendor issues none. The recorder is a fixture scoped to this suite, and instance-queue machinery for an unreachable case is speculative weight. If a future test asserts on a redirecting path, that's the commit to add it in.

Issue #240, opened 2026-08-28
generate cli: OpenAPI-sourced generation and an importable extension module


DonaldMurillo:
## Motivation

`gofastr generate cli` emits a terminal client for the app's HTTP API, but it generates from the entity set. An app with a hand-written API has no entities, so the generator has nothing to read and the app gets no CLI.

Concrete case: gobarc (barcode.donaldmurillo.com). Its API is five hand-written endpoints (`/api/generate`, `/api/validate`, `/api/decode`, `/api/batch`, `/api/spec`) and it now publishes an accurate, drift-tested OpenAPI 3 document at `/openapi.json` — authored by hand for the same reason: no entities to generate from. The is-agentic scanner's "CLI tool available" check fails for exactly this class of app, and the v0.72.0 `AgentReadyConfig.CLI` llms.txt section has nothing to advertise.

Two asks, which compose:

## 1. OpenAPI as a generation source

```
gofastr generate cli --from-openapi openapi.json [--out cmd/<binary>]
```

Generate the same style of client from an OpenAPI 3 document instead of the entity set:

- one subcommand per operation, named from `operationId` (fail on missing/duplicate ids rather than inventing names, matching the existing no-auto-renaming stance on flag collisions)
- typed flags from `parameters` and the request-body schema; the existing scalar mapping (string/int/float64/bool, `map[string]any` for open objects) already covers it
- binary request bodies (e.g. an image upload for a decode endpoint) map to a file-path flag or stdin
- same conventions as the entity CLI: `--url`/`--token`, env prefix, `-o`/`--json`, one-shot owned code, `custom.go` never overwritten

The spec can be a local file or a URL. Apps that generate their spec keep using entities; apps that hand-author a spec get the CLI from the artifact they already maintain.

## 2. An importable extension module, for any base generation

`custom.go` is a single file inside the generated package. That works for small tweaks, but extensions can't be versioned separately from the generated tree, shared between apps, or survive a `--out` move.

```
gofastr generate cli --extend github.com/me/myapp/clix
```

The generated `main.go` imports the package and merges its commands over the base set with the same semantics `customCommands()` has today: matching names replace, new names add, wrapping calls the generated run function. Something like:

```go
// package clix — developer-owned, outside the generated tree
func Commands(c *client.Client) []cli.Command { ... }
```

[...]

PR #264, opened 2026-08-28
feat(cli): generate cli --from-openapi (#240)

DonaldMurillo:
Implements the accepted half of #240 per the design posted there (the `--extend` half stays deferred on the issue).

```bash
gofastr generate cli --from-openapi openapi.json --binary=barc
gofastr generate cli --from-openapi https://barc.example.com/openapi.json
```

One subcommand per `operationId` — missing or duplicate ids fail generation with the path+method named, matching the entity CLI's no-auto-renaming stance. Path/query/header parameters become typed flags (string/int/float64/bool; arrays of scalars repeat the flag); `application/json` object bodies get a flag per top-level scalar property plus the `--json` escape, with spec-`required` fields enforced; binary bodies take `--file` (`-` = stdin). JSON responses pretty-print, anything else streams raw to stdout so `barc generate-code ... > code.png` works. `servers[0].url` seeds the default URL; `http: bearer` and header `apiKey` schemes ride the existing `--token`/env/`login` machinery (the apiKey header name comes from the scheme).

The generated tree is self-contained: `internal/client` mirrors the entity SDK's surface (`NewClient`, `Token`, `APIError`, `Do`), so all the shared scaffolding — config storage, `login`/`logout`, `custom.go`, `--force` semantics — is byte-identical between the two flavors, and the output builds with zero framework dependency (it runs from any Go module, no entities package required). Supported subset is explicit and enforced: local `#/components/...` refs only, `allOf` shallow-merge, `oneOf`/`anyOf` bodies fall back to `--json`; cookie parameters, external refs, multipart, and OAuth2 flows fail with precise errors rather than guessing. YAML documents parse through `core/yaml` — no new dependencies.

This also closes the is-agentic gap from the issue's motivation: an app that hand-authors its spec gets a CLI from the artifact it already maintains, so `AgentReadyConfig.CLI` has something real to advertise.

Testing: spec-model unit tests (typing, ordering, security mapping), five error-path tests (missing id, duplicate command, reserved flag, external ref, YAML), a generated-tree `go build`, and a live `httptest` round trip proving path escaping, repeated query params, bearer auth, body-from-flags, required-field enforcement, and a binary upload against a real listener. The duplicate-detection guard was mutation-proven (check disabled → test red → restored). Docs in `app-cli.md` ship in this commit; the doc compile gate passes.

## Summary by CodeRabbit

- **New Features**
  - Generate standalone terminal clients from OpenAPI 3 documents using a local file or URL.
  - Create operation-based subcommands with typed parameters, JSON and binary request support, and raw response streaming.
  - Support default server URLs plus bearer-token and API-key authentication.
  - Accept OpenAPI JSON or YAML documents and produce a self-contained client.

- **Documentation**
  - Added usage guidance, supported features, and generation limitations to the CLI documentation.


DonaldMurillo:
All six CodeRabbit findings plus the adversarial-review findings are addressed in 19cf8a5f, none rejected:

- **--json + query/header flags (Major)**: body construction is now runner-local; exclusivity checks body field flags only. Live round-trip test added (`--json` + `--dry`).
- **Array path params (Major)**: rejected with a precise error, as are typed (non-string) arrays anywhere — repeatable flags collect strings, so typed arrays would have skipped validation silently.
- **Go identifier sanitization (Minor)**: derived names are validated against a strict grammar and rejected rather than munged (no-auto-renaming stance); this also closes the source-injection route for URL-fetched specs, together with %q on the placeholder/usage interpolations.
- **binary/fromOpenAPI header validation (Minor)**: control-character check applied to both before they reach the generated header comment.
- **Test race + partial body read (Minor ×2)**: mutex-guarded capture and io.ReadAll.

Beyond the line comments: entity-mode reserved words (limit/page/sort/q) no longer block generation — an OpenAPI parameter name is the wire name — and $ref cycles, path/template mismatches, degenerate operationIds, templated server URLs, nameless apiKey schemes, and requestBody.required all fail or enforce loudly now. Fetch has a 30s timeout.

Issue #278, opened 2026-08-28
flaky: TestE2E_RuntimeSplit_ToastModuleFailureShowsFallback times out under CI load


DonaldMurillo:
Failed once on CI (run 33210798073, PR #276 — a docs-only diff, so the test's own inputs were unchanged), passed on rerun and passes locally at -count=3. Failure mode:

```
--- FAIL: TestE2E_RuntimeSplit_ToastModuleFailureShowsFallback (10.66s)
    e2e_runtime_modules_test.go:374: X-Gofastr-Toast header fired with toasts.js returning 500 should render a fallback notice carrying the server-sent title; nothing visible was found.
```

The ~10.6s duration suggests the fallback-notice wait hit its deadline under runner load rather than the fallback never rendering. Same chromedp flaky-under-load class the suite has hardened against before (shared browser, serialized jobs). Candidate fixes: lengthen/poll the fallback-visibility wait the way the suite's other slow-path assertions do, or assert on a deterministic DOM signal the runtime sets before the visual notice.

Filing after one occurrence because the retained failed attempt made a merged PR (#276) look red in the commits view — worth killing the flake rather than re-rolling it. One data point so far; if it never recurs after the wait hardening, close.

DonaldMurillo:
Closing per the issue's own criterion. The wait hardening landed in #285 (`examples/site/e2e_runtime_modules_test.go`, +26/-8) and the failure has not recurred on any CI run since. Reopen if it comes back.

DonaldMurillo:
Reopening: this recurred, so my close was premature.

I closed this earlier today on the issue's own criterion — hardening landed in #285 and it had not recurred. It has now, on PR #308 (run 33277533905, `browser e2e · site`), on a diff that touches only a `framework/` test file and cannot reach `examples/site`.

**The new diagnostic points somewhere different from the original load-under-CI hypothesis:**

```
fallbackPresent: false, fallbackText: null
loadedModules: ["toasts", "copy", "disclosure", ...]
runtimeScripts: [".../toasts.js?v=e26313f0407f2ddb", ...]
```

The test forces `toasts.js` to return 500 and expects a fallback notice. But `toasts` appears in **`loadedModules`** and its script in `runtimeScripts` — the module loaded successfully. So this run did not fail because the fallback was slow to appear; it failed because **the 500 injection did not take effect**, and there was correctly no fallback to find.

[...]

PR #285, opened 2026-08-29
test(site): harden the toast-fallback e2e wait against runner starvation

DonaldMurillo:
Hardens the wait in `TestE2E_RuntimeSplit_ToastModuleFailureShowsFallback` per #278. Not marked as closing the issue: it stays open until the flake stops recurring, per its own filing note.

## What changed

- The fallback-visibility poll budget goes from 10s to 30s. The failed CI attempt ran 10.66s — the poll expired rather than erroring. Module scripts load with `async=false`, so the toasts.js error event that triggers the fallback queues behind every earlier in-order boot script; under runner starvation that chain can outlive 10s. The per-test tab context is 45s, so 30s fits.
- A timeout now resolves a diagnostic snapshot (fallback node state, runtime script srcs, loaded modules) instead of a bare `false`, so a recurrence reports what the page looked like instead of "nothing visible was found".
- The poll promise is pinned as `window.__toastFallbackPoll`. At the longer budget Chrome GC collected the awaited promise (`Promise was collected` -32000) because CDP holds it only weakly once the tick chain stops; the pin keeps the diagnostic deliverable. (This GC failure can't be the original CI flake — a collected promise fails the test via `chromedp:` fatal, not the observed `t.Errorf`.)

## Verification

- `go vet ./examples/site/` clean; the test passes at `-count=3` (~1–2s per run).
- The timeout path was proven by mutation: predicate broken, run observed failing with the diagnostic JSON delivered intact, predicate restored. The captured diagnostic also confirmed the in-order-queue analysis — toasts.js sat behind 8 other boot-queued modules.

No runtime or framework code changed; test-only.

https://claude.ai/code/session_01MnA5y1Dynm4T1AGvAMbPMe

### DIFF
--- a/examples/site/e2e_runtime_modules_test.go
+++ b/examples/site/e2e_runtime_modules_test.go
@@ -305,7 +305,7 @@ func TestE2E_RuntimeSplit_ToastModuleFailureShowsFallback(t *testing.T) {
 	base := srv.URL
 	ctx := newE2EBrowserCtx(t)
 
-	var fallbackVisible bool
+	var fallbackResult string
 	// Navigate to a page that doesn't pre-load the toast module so
 	// toasts.js isn't already cached, then manually trigger the
 	// toast push path to exercise the fallback when the module 500s.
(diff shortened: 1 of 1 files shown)


DonaldMurillo:
@coderabbitai review

Issue #279, opened 2026-08-28
battery/admin: destructive module actions have a confirm attribute that never fires


DonaldMurillo:
The process-modules admin screen puts `data-fui-confirm` on its destructive submit buttons (`battery/admin/process_modules.go` — capability revoke, and the action forms built by `moduleActionForm`). The runtime only reads that attribute in `_dispatchRPC` (`core-ui/runtime/src/rpc.js:127`), which requires `data-fui-rpc` on the node. These are plain POST forms, so the confirm has never fired: revoking a capability or restarting a module executes on first click.

Found while migrating ExtraAttrs sites for #262. The attribute still renders (ui.Button is a wiring carrier), it just does nothing.

Two ways to fix, pick one:

1. Teach the runtime to honor `data-fui-confirm` on the submitter of an intercepted plain form submit. New runtime semantics — needs the runtime test suite and a runtime-contract.md note per the data-fui-* rule.
2. Convert the admin forms to RPC (`interactive.OnSubmit` + `WithConfirm`). The handlers 303-redirect to the list, so the RPC path's redirect handling needs verifying end-to-end.

Either way, an e2e test should click revoke and assert the confirm gate, since the current gap was invisible to every existing test.

DonaldMurillo:
Fixed in #293 (merge c0266af3). The gate now runs in the submit bridge before every branch, so native POST, data-fui-spa, and data-fui-rpc forms are all covered, and the widget-scoped listener matches. A submit button's message takes precedence over the form's, which is why battery/admin needed no change - both call sites already carried the attribute on the button. Review caught that form-level placement was pinned by nothing once the bridge started passing opts.confirmed; TestConfirmFormLevelMsgGates closes that, mutation-proven (dropping the form fallback fails only that test and leaves the other three green).

Issue #255, opened 2026-08-28
pluginhost: no opt-in tier for wasm in the framed CSP, so no plugin can run a WebAssembly engine


DonaldMurillo:
`framedCSP` in `framework/pluginhost/assets.go` is a fixed string with no per-plugin extension point. It has no `'wasm-unsafe-eval'`, so WebAssembly cannot instantiate inside a plugin frame at all.

This is not hypothetical. The pdf plugin in gofastr-plugins runs pdf.js **worker-free on the main thread** specifically to avoid the wasm paths, and that workaround has stood in for an answer ever since. Any plugin built on a wasm engine — DuckDB for a SQL notebook, zxing for a barcode scanner, ONNX for local inference — is currently impossible in the sandbox, and the only escape is to make it a trusted host-page plugin, which throws away the isolation the project exists to demonstrate.

## What the tier would be

A per-manifest CSP extension validated against a closed allowlist whose only member is `wasm-unsafe-eval`:

```go
// Manifest
CSP []string `json:"csp,omitempty"`
```

`Validate()` rejects any token outside the allowlist, so a plugin cannot smuggle in a general CSP escape — not `'unsafe-eval'`, not extra hosts, not arbitrary sources. `framedCSP` appends the quoted keywords to `script-src` and changes nothing else.

## What must not change, and why the tier is narrow

The opaque origin, `sandbox allow-scripts`, `connect-src 'none'`, and the `allow-same-origin` ban all stay exactly as they are. `wasm-unsafe-eval` permits WebAssembly compilation inside `script-src`; it does not enable string eval and does not touch network or storage. A frame running DuckDB under this tier still cannot fetch, cannot reach host cookies or DOM, and cannot exfiltrate — the data has to arrive over the postMessage bridge and the answers have to leave the same way.

That is a genuinely interesting property worth having: a database engine running inside a cage that cannot phone home.

## Caveat worth settling in the same change

DuckDB's multi-threaded build wants Web Workers and `SharedArrayBuffer`, which need COOP/COEP cross-origin isolation and fight the opaque-origin design head-on. The realistic answer is that only single-threaded wasm builds work under this tier, and that limit should be documented rather than discovered.

## Blocking

gofastr-plugins #21 (`sqlnotebook`) is blocked on this and is the intended first consumer. Filing rather than patching, per the rule that gofastr changes go through issues.

DonaldMurillo:
[...]

Issue #252, opened 2026-08-28
pluginhost: no frame→host request/response channel, so every plugin hand-rolls correlated events


DonaldMurillo:
`framework/pluginhost/host/pluginhost.js` gives the host a `request(method, params)` that returns a promise resolved by a matching id. That is **host → frame** only.

Frame → host has no equivalent. A plugin that needs to ask the host something and wait for the answer has to invent its own correlation, and every plugin invents the same one:

- richtext: `requestUpload` → `uploadResult` (`richtext/host/broker.js`, `richtext/js/src/editor.ts:174`)
- datagrid: `requestRows` → `rowsResult`, `requestCellWrite` → `cellWriteResult`, `requestExport` → `exportResult` (`datagrid/js/src/bridge.ts`)

Both hand-roll a `reqId`, a pending map, and their own timeout policy. The sqlnotebook plugin is about to be the third, and imageedit will be the fourth.

## Why it matters beyond duplication

The pieces each plugin reimplements are exactly the ones that are easy to get subtly wrong: request timeouts, an unmatched response, a response arriving after teardown, and a bounded pending map. The host side already solved all four for its own direction.

## What would fix it

A `sendRequest(method, params, timeoutMs)` on the frame client that mirrors the existing host-side `request()`, with the host adapter registering handlers by method name. The wire shape does not need to change: it is the same envelope with an id, in the other direction.

Found while building the datagrid plugin in gofastr-plugins, where the paging round trip is the plugin's core mechanic.

DonaldMurillo:
Three-model grooming: **accepted — with the design scoped wider than the ticket**. Verified: host→frame `request()` exists with pending map + timeout (pluginhost.js:284-315) and inbound dispatch handles only `response`/`event` (:398-409), so a frame-sent `request` is silently dropped today — which conveniently makes the addition forward-compatible with old brokers (frozen envelope, unknown types ignored).

Two findings that reshape the work: the existing host side has **no pending-map bound and does not reject outstanding promises on teardown** (:432-446) — it has not actually solved all four failure modes the ticket credits it with, so mirroring it verbatim would copy those gaps into the new direction. And the framework ships no reusable frame client for a `sendRequest` to live on.

[...]

Issue #281, opened 2026-08-28
ui.Button: aria-label via ExtraAttrs has always been dead — repeated buttons announce identically


DonaldMurillo:
`ui.Button` derives `aria-label` from Label and always has: `html.Button` re-asserts the Label-derived value after copying extras (`core-ui/html/interactive.go`), so an `aria-label` passed through Button ExtraAttrs never rendered. PR #280 makes that explicit (Button drops `aria-label` as an owned key), but the sites that tried it still have the underlying accessibility problem: a list of visually identical buttons is announced identically.

Known sites:

- `battery/admin/rbac_admin.go:149` — every row's revoke button announces as "Revoke" with no role/user context.
- `examples/site/screen_livedash.go:472,480,488` — three dashboard buttons with the same label; `examples/site/e2e_livedash_test.go:268` documents the clobber instead of fixing it.

Fix direction: ButtonConfig needs a real way to disambiguate — either an `AriaLabel` config field that wins over the Label-derived default, or the call sites should put the context in the visible Label. Whichever way, the e2e test that documents the clobber should start asserting distinct accessible names.

Found by the PR #280 review (finding 3); pre-existing, not a regression.

DonaldMurillo:
Shipped in #288 (merge 34300c92). ButtonConfig.AriaLabel on both html.Button and ui.Button sets the accessible name independently of the visible Label; ui.Button still strips an ExtraAttrs aria-label as an owned key. The icon-only panic guard was ignoring AriaLabel — a real bug found while migrating call sites, fixed and mutation-proven. Migrated the admin RBAC revoke buttons and the three livedash incident buttons, and the livedash e2e now asserts the rendered aria-label rather than only claiming to.

PR #293, opened 2026-08-29
batch: wasm CSP tier (#255) + data-fui-confirm on plain form submits (#279)

DonaldMurillo:
Batch 2, closing #255 and #279 (and #252, the framed-runtime umbrella, whose
third and last PR this is).

## #279 — `data-fui-confirm` fires on plain form submits

The ticket read as an admin-battery bug. It isn't: the attribute was only ever
read in `_dispatchRPC`, which needs `data-fui-rpc` on the node, and a plain POST
form leaves the submit bridge at the enctype check (`frag/boot.js:270`). The gate
has never run for **any** app that put it on a plain form — the admin screen just
noticed first, revoking a capability and disabling a module on the first click.

The gate now runs before every branch of the submit bridge, so native,
`data-fui-spa`, and `data-fui-rpc` forms are all covered, and the widget-scoped
listener gets the same treatment. A submit button's message wins over the form's,
because one form can carry several submit buttons of different destructive
weight — exactly the shape the admin screen uses, which is why `battery/admin`
needed no change. Both dispatch sites pass `{confirmed:true}` so `rpc.js` doesn't
prompt twice; the click path keeps its own prompt.

**Budget note, flagged for review:** the core gzip line moves 14700 → 14784, and
the previous comment in `budget_test.go` warned the next feature should carve
rather than raise. The gate can't be carved into a demand module — a native
submit navigates away before one could load — so the 85 bytes have to sit in
core. It clears the real bundle by 39 bytes, which is tight.

## #255 — opt-in wasm tier for the framed CSP

`framedCSP` was a fixed string with no `'wasm-unsafe-eval'`, so WebAssembly
couldn't instantiate in a plugin frame at all. `Manifest.CSP` opts in against a
closed allowlist with exactly one member, widening `script-src` and nothing else.
The opaque origin, `sandbox allow-scripts` without `allow-same-origin`, and
`connect-src 'none'` all stay, so a wasm engine still can't fetch, reach host
cookies or DOM, or exfiltrate.

Matching is exact, byte-for-byte. The sandbox validator normalises case and
whitespace because the HTML sandbox attribute tokenises that way, making a
variant there a live grant; the CSP header doesn't normalise source expressions,
so `'WASM-UNSAFE-EVAL'` is inert rather than an evasion, and one comparison
rejects every smuggle shape — including a token carrying `;`, which could
otherwise splice a directive into the response header and re-enable
`connect-src`. The assembly point re-filters through the same allowlist, so a
slice that skipped `Validate` can't smuggle a keyword either.

Only single-threaded wasm builds work under this tier; multi-threaded builds want
`SharedArrayBuffer` and COOP/COEP cross-origin isolation, which fight the opaque
origin. Documented rather than left to be discovered.

## Verification

Every new guard was mutation-proven red→green. I re-ran the ones I cared most
about myself rather than trusting the report: disabling the boot.js gate turns
`TestConfirmNativeFormCancelBlocks` red, and restoring it green.

The wasm proof drives a real Chrome: the tiered frame instantiates a minimal
8-byte module, and the control frame fails with Chrome's own CSP error. The test
asserts the failure *is* the CSP gate, not merely that it failed, so it can't
pass vacuously.

**One caveat worth stating plainly:** that proof sits behind `//go:build
chromium`, and CI never passes that tag (ci.yml says so deliberately). So it is a
manual proof, not an enforced gate. What CI does protect is the header assembly —
`script-src`-only and byte-identity for a manifest without the tier, both
mutation-proven. The browser test only regresses if Chrome changes rather than if
we do, which is why I left it where the repo puts its other browser-level proofs.

Suites run: `core-ui/runtime` isolated (240s, green), `framework/pluginhost`,
`go build ./...`, `go vet ./...`.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in WebAssembly tier for framed plugins through a validated manifest setting.
  * WebAssembly remains blocked by default and is enabled only with the approved security token.

* **Bug Fixes**
  * Confirmation prompts now apply to native, SPA, and RPC form submissions.
  * Submit-button messages take precedence over form messages.
  * Canceling a confirmation now reliably prevents submission, with no duplicate prompts.

* **Documentation**
  * Updated plugin and interactive-pattern guidance for WebAssembly and confirmation behavior.


DonaldMurillo:
@coderabbitai review

DonaldMurillo:
## Independent review record

CodeRabbit is rate-limited on this repo (free OSS quota exhausted); it
acknowledged a `@coderabbitai review` trigger but has not delivered one, and
Copilot cannot be added as a reviewer (not a collaborator). So this PR would
otherwise merge with **zero** bot reviews — and the review gate reports
"0 unresolved" in that state, which is indistinguishable from a clean review.
Recording a manual review here rather than merging on a vacuous green.

### Guards re-proven independently

Not taken from the worker reports — re-run here:

1. **Native-form confirm gate.** Disabling the boot.js gate
   (`if (false && msg && …)`) and regenerating `runtime.js` turns
   `TestConfirmNativeFormCancelBlocks` red; restoring it, green.
2. **CSP allowlist re-filter at assembly.** The worker's mutation table
   credited one test with catching this, but the same edit also breaks dedupe —
   two different properties behind one assertion. Dropping *only* the allowlist
   half (keeping `seen[kw]`) lets `'unsafe-eval'` reach `script-src`:
   ```
   got  "… script-src http://h 'wasm-unsafe-eval' 'unsafe-eval'; …"
   want "… script-src http://h 'wasm-unsafe-eval'; …"
   ```
   So the defence-in-depth claim ("a slice that skipped `Validate` cannot
   smuggle a keyword") is genuinely tested, not incidental.
3. **Browser proof.** Run locally: the tiered frame instantiates an 8-byte wasm
   module, the control frame fails with Chrome's own CSP error. The test asserts
   the failure *is* the CSP gate, not merely that it failed, so it cannot pass
   vacuously.

### Anti-vacuity check on the new tests

`TestConfirmNativeFormCancelBlocks` asserts the **handler-hit count**, not just
that `window.confirm` fired — counting prompts alone would pass even if the form
submitted anyway — and sleeps 700ms first so an ungated submit would have landed.

### Finding, fixed in this PR

`Manifest.CSP` was declarative-only. Nothing in this repo builds an
`AssetServer` from a `ClientModule` (the registration path lives in
gofastr-plugins), so declaring `csp` and forgetting `.WithCSP(...)` validates
cleanly and still refuses WebAssembly, with nothing reporting why. That is the
third silent-no-op of the same class in this batch — the dead `aria-label`
(#281) and the confirm that never fired (#279) are the other two — so it is now
stated at the field a plugin author reads. Doc sentence rather than a
`NewModuleAssetServer` helper, because the standing call on this project is to
stop adding API surface.

### Known limitation, stated rather than glossed

The wasm browser proof sits behind `//go:build chromium`, and CI never passes
that tag (ci.yml says so deliberately). It is a manual proof, not an enforced
gate. What CI protects is the header assembly — `script-src`-only plus
byte-identity for a manifest without the tier, both mutation-proven above. The
browser test only regresses if Chrome changes rather than if we do, which is why
it sits where the repo keeps its other browser-level proofs.

DonaldMurillo:
GLM review (omp) run in place of CodeRabbit, which is out of free OSS reviews. No security defects: it traced every smuggling shape against `framedCSP`, confirmed the sink filter is authoritative for hand-built manifests that skip `Validate`, audited all five entry paths into `dispatchRPC` for double/missed prompts, and independently ran the chromium wasm proof and the full 244s runtime suite.

Three Minor findings, all verified against the code and all fixed in 06cc1581:

1. **Form-level `data-fui-confirm` placement was pinned by nothing.** Every fixture put the attribute on the submit button. Before this PR the form-level read was covered incidentally by `_dispatchRPC` reading the node's own attribute; the bridge now passes `opts.confirmed`, which skips that read. The reviewer predicted that narrowing the gate to the submitter alone would keep the suite green — confirmed by mutation: dropping the form fallback fails only the new `TestConfirmFormLevelMsgGates` and leaves the other three confirm tests passing. That is the coverage hole closed.
2. `framework/docs/content/runtime-contract.md:109` still described the old RPC-only semantics. That is the copy `gofastr docs` serves, and it contradicted the internal contract on the same row.
3. `ui.MenuItem.Confirm`'s doc comment claimed the runtime reads the attribute only on RPC dispatch.

Rejected nothing as invalid. Not acted on, deliberately: a CI matrix shard for the chromium-tagged wasm proof (real gap, but belongs with a CI change rather than this batch), nested-widget double-dispatch (pre-existing, not introduced here), and `e.submitter` on old Safari (degrades to the form-level message, which now has a test).

PR #324, opened 2026-08-30
blueprint: derive marketing links and the auth gate from registered screens

### LINKED ISSUE #312: Blueprint marketing chrome hardcodes /pricing, /about, /terms, /privacy — routes the generator never registers
`cmd/gofastr/blueprint.go` emits the marketing header and footer with four literal hrefs:

- `:7019`, `:7043` — header nav: `/pricing`, `/about`
- `:7054`–`:7056` — footer: `/pricing`, `/about`, `/terms`, `/privacy`

That chrome is emitted whenever `hasMarketing` is true, which is set by any screen declaring `layout: marketing` (`:6871`–`:6878`). The screens themselves come from `bp.Screens` via `blueprintScreenMountStmt` (`:4917`–`:4939`). Nothing connects the two. A blueprint with one marketing screen and no `/terms` screen generates an app whose footer links to a 404.

`examples/site/exampleblueprints/meridian.yml` declares all four routes (`:298`, `:335`, `:350`, `:358`), so the one shipped example passes. That is a coverage gap, not evidence the code is right.

## Already fixed once, for one link

The comment at `:7025` documents the same bug class, found and fixed for `/login`:

> blueprintLoginRoute, not a hardcoded "/login": the app-shell sidebar had the same literal and it pointed at a route six of the seven shipped blueprints never register. Dormant here today only because meridian is the sole blueprint with both marketing screens and auth, and it happens to route /login, which is luck, not design.

`blueprintLoginRoute` derives that one link from the blueprint. The four literals directly below it were left alone.

## Fix

Either:

1. Derive each nav/footer entry from `bp.Screens` and drop the ones no screen registers. Smaller change, and it matches what `blueprintLoginRoute` already does.
2. Generate the missing screens.

Option 2 overlaps the scaffolding half of the launch-checks proposal and should be decided there.

## Test

No test compares generated hrefs against the generated route table. A regression test should generate from a blueprint with `layout: marketing` and no `/terms` screen, then assert every emitted href resolves to a registered route.

DonaldMurillo:
Closes #312, and fixes the higher-severity sibling found while doing it.

## The filed bug

The marketing header and footer carried four literal hrefs — `/pricing`,
`/about`, `/terms`, `/privacy` — emitted whenever any screen declares
`layout: marketing`. Nothing connected them to `bp.Screens`, so a blueprint
with one marketing screen and no `/terms` screen generated an app whose footer
linked a 404.

This bug class had already been found and fixed **once, for one link**. The
comment sitting directly above these literals explains that `blueprintLoginRoute`
derives the login href because the hardcoded one pointed at a route six of the
seven shipped blueprints never register. The four literals below it were left
alone.

They now come from one table filtered by the routes `bp.Screens` registers,
shared by header and footer — two independent lists is how they drifted in the
first place. A footer column with no surviving links is omitted rather than
emitted empty, since `ui.SiteFooter` renders a present column as a title over
an empty `<ul>`.

**meridian generates byte-identical chrome**, which is the point rather than a
caveat: it declares all four routes and is the one shipped blueprint that
*cannot* observe this bug. The new test uses a fixture that can.

## The sibling, which is worse

`authPolicy` emits a **redirect** for every unauthenticated visitor to a gated
screen, and both emitters — the screen mount and the island policy — hardcoded
`"/login"`. A blueprint whose login screen sits at `/signin` generated an app
where every gated page redirects to a route it never registers. That is not a
dead footer link; it is an app with no working sign-in path.

`bp` is threaded rather than a login-path string. Both sites need it several
layers down, and a bare `string` parameter next to the existing `apiBase string`
is a swap the compiler cannot see — a distinct type makes that mistake a build
error instead of a silently wrong redirect.

## A near-miss worth recording

My first fixture gated a screen with a `hero` block. That emits no island
policy, so restoring the island literal changed nothing and **the mutation
passed** — the guard looked proved while half of it was untested. The fixture
now carries an `entity_list` block on the gated screen, and the test fails if
the generation stops emitting *either* gate, so that half cannot go quiet
again.

Both literals are mutation-proved:

```
[A: mount-site literal restored]   FAIL
[B: island-policy literal restored] FAIL
```

## Still-literal routes, reported not fixed

Found in the same emitter, left alone deliberately:

- `Href: "/"` for the header brand and footer lead, and `SignOut{Next: "/"}`.
  Same class, but `/` is the conventional target and a marketing blueprint
  without a home screen is a different conversation.
- `blueprintLoginRoute` itself falls back to `/login` when no screen hosts a
  login form, so an auth app with no login screen still links to an
  unregistered page. That is the established helper's own default; changing it
  is a design call, not a bug fix.

## Verified

`go build`, `go vet`, `gofmt` clean. `go test ./cmd/gofastr/ -count=1` green
(271s), including the generated-app gates and the meridian e2e generation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,19 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   supported way to attach attributes to a component — cannot reach it.
   Inert for everyone else.
 
+### Fixed
+
+- **Generated marketing chrome links and auth-gate redirects follow the
+  blueprint's registered screens** (#312) (`gofastr generate`). The marketing
+  header nav and footer shipped four literal hrefs (`/pricing`, `/about`,
+  `/terms`, `/privacy`), so a marketing blueprint without those screens got a
+  footer full of 404s; a chrome link is now emitted only when a screen
+  registers its route. The auth gate's redirect — the screen mount, the
+  entity-list island policy, the header's Sign in button, and the failed-login
+  bounce — derives from the screen hosting the login form (nested in a section
+  counts) instead of the hardcoded `/login`, which 404'd every gated page on a
+  blueprint whose sign-in lives elsewhere.
+
 ## [0.76.0] - 2026-08-30
 
 ### Added
--- a/cmd/gofastr/blueprint.go
+++ b/cmd/gofastr/blueprint.go
@@ -1085,6 +1085,47 @@ func blueprintAppHome(bp Blueprint) string {
 	return "/"
 }
 
+// blueprintMarketingLink is one entry of the marketing chrome's canonical
+// link table. The header nav and the footer columns both project out of it,
+// so the two surfaces cannot drift apart again.
+type blueprintMarketingLink struct {
+	Label  string
+	Route  string
+	Nav    bool   // included in the header nav when its route is registered
+	Column string // footer column title when registered; "" = not in the footer
+}
+
+// blueprintMarketingLinkTable is the single source of truth for the marketing
+// chrome's links. A link is emitted only when a screen registers its route —
+// the rule blueprintLoginRoute established for "/login". Before that, the
+// chrome shipped four literal hrefs, so a one-screen marketing blueprint
+// generated a footer full of 404s (#312).
+var blueprintMarketingLinkTable = []blueprintMarketingLink{
+	{Label: "Pricing", Route: "/pricing", Nav: true, Column: "Product"},
+	{Label: "About", Route: "/about", Nav: true, Column: "Company"},
+	{Label: "Terms", Route: "/terms", Column: "Legal"},
+	{Label: "Privacy", Route: "/privacy", Column: "Legal"},
+}
+
+// blueprintRegisteredMarketingLinks returns the canonical marketing links
+// whose routes bp.Screens actually registers, in table order (which is also
+// the footer's column order). Screens with path params mount at the
+// colon-style route, so the same normalization blueprintScreenMountStmt uses
+// applies here.
+func blueprintRegisteredMarketingLinks(bp Blueprint) []blueprintMarketingLink {
+	registered := make(map[string]bool, len(bp.Screens))
+	for _, s := range bp.Screens {
+		registered[blueprintScreenRoutePath(s.Route)] = true
+	}
+	var out []blueprintMarketingLink
+	for _, l := range blueprintMarketingLinkTable {
+		if registered[l.Route] {
+			out = append(out, l)
+		}
+	}
+	return out
+}
+
 // screenHasAuthForm reports whether a screen hosts a login or signup form, so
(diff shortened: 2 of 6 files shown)


DonaldMurillo:
@coderabbitai review

No review coverage yet. Two things worth attacking specifically: (1) the marketing-link filter — can a blueprint produce chrome that links a route it does not register, or conversely lose a link it should keep; (2) the auth-gate change, which threads `bp` through several emitters to derive the login redirect. The second is the security-relevant half: getting it wrong means every gated page redirects somewhere the app never mounts.

DonaldMurillo:
## Independent review: no production defect, two proven holes in the guards

The review verified the part I most wanted checked — that a wrong-but-compiling `bp` at any of the ~30 threaded call sites would silently restore `/login`. It enumerated every one, confirmed the two `Blueprint{}` test sites are harmless (neither exercises a path that consults `bp`), and found no remaining `authPolicy(` literal anywhere.

It also verified byte-identical output rather than trusting my claim: built both binaries, generated **all seven** shipped blueprints, `diff -r` on each pair — identical.

Then it found two holes in the guards, both proven by full-suite-passing mutations. Fixing all of it; the fixes are in flight.

### 1. The marketing test cannot catch over-filtering

It asserts only `emitted hrefs ⊆ registered routes` — never the converse. `blueprintRegisteredMarketingLinks → return nil` **passes the entire suite**. So does inverting the nav builder's flag, because header and footer hrefs are pooled into one list and the footer still supplies `/pricing`.

That is precisely the regression the test exists to prevent: a future normalisation applied to one side only drops Pricing and About from every marketing app, and CI stays green. Positive per-surface assertions going in.

### 2. The nested `entity_list` call site has zero coverage

Replacing the single `bp` at `blueprint.go:6257` with `Blueprint{}` — the exact threading bug this PR is most at risk of — **passes the full suite, 261s, exit 0**. No fixture combines an auth-gated screen with an `entity_list` nested in a section child, so that call site's derivation is never observed. Fixture being reshaped to cover it.

### 3. My commit message is false for one blueprint shape

`blueprintLoginRoute` scans `s.Body` flat, so a login form nested inside a `section` is invisible to it and the gate falls back to `/login`. Not a regression — the old literal produced the same 404 — but the commit claims the gate points at the blueprint's login route, and for that shape it does not. `screenHasAuthForm` has the same flat scan, so such a screen also misses `guestPolicy` and shows the login form to signed-in users.

Recursing both scanners into `block.Children`, following the walker at `:4114` that already does it.

### Deliberately not in this PR

The `Href: "/"` brand and `SignOut{Next: "/"}` literals, and trailing-slash/case route normalisation. Both real, both pre-existing on main, and both a different change — the second in particular is a decision about whether to *fix* an author's typo or *hide* the link, which deserves its own discussion rather than riding along.

Worth noting what the review's shape says about mine: I checked that the threading compiled and that the tests passed, and both were true. What I did not check was whether the tests could fail — which is where both real findings were.

Issue #319, opened 2026-08-30
ui.Menu cannot express submenu + menuitemradio contracts (nested role=menu, aria-haspopup items, id-addressable rows)


DonaldMurillo:
Porting metacollector's dashboard header onto ui.Menu hits three gaps; the suite pins this contract (tests/customization/theme.spec.ts, tests/dashboard.spec.ts, tests/help.spec.ts, tests/imports.spec.ts).

1. **No submenus.** The user menu nests a Palette submenu: an item with `aria-haspopup="menu"` revealing a child `role="menu"` panel. MenuConfig has no Children on MenuItem, so hosts with that shape can't adopt the component.

2. **No menuitemradio.** Theme selection renders `role="menuitemradio"` rows with `aria-checked` (specs click `getByRole('menuitemradio', { name: /dark/i })`). MenuItem only emits menuitem shapes; a radio group (with a group name) is inexpressible.

3. **Item ids are dropped.** MenuItem.ExtraAttrs explicitly drops `id`, so rows that host runtimes address by id (e.g. a Help Mode toggle the page's JS binds to, an Imports & Exports row that opens an activity sheet) can't survive the migration. An explicit ID field (or allowing id in ExtraAttrs) would unblock id-pinned contracts.

Suggested shape, mirroring how Sidebar gained GroupMarkup: `MenuItem{Children []MenuItem, Radio string, ID string}` — a Radio group name renders `menuitemradio` rows with aria-checked; Children render the nested panel with the same disclosure machinery; ID passes through to the row element.

The port (go-collect-fastr) is the acceptance test — theme/help/imports specs exercise all three.

DonaldMurillo:
## Scoping this before it gets written, since it is three changes wearing one ticket

#320 shipped (#326). This one is deliberately not batched with it: #320's items were attribute additions, whereas all three of these change `MenuItem`'s **structure**, and they do not share a mechanism with each other either.

**1. Submenus** — nested `role=menu` behind an `aria-haspopup="menu"` item. This is the expensive one: it needs a nesting model, keyboard semantics (right/left to open/close, roving focus across levels), and a decision about whether the nested panel reuses the existing disclosure machinery or gets its own. It also has an open a11y question the ticket does not settle — whether the parent item stays activatable or becomes purely a disclosure.

[...]

PR #326, opened 2026-08-30
ui.Tabs: data-state, aria-controls wiring, and vacated hidden panels

### LINKED ISSUE #320: ui.Tabs cannot express data-state/aria-controls contracts or vacate hidden panel content
Porting metacollector's tabstrips onto ui.Tabs hits a contract gap; the suite pins this shape (tests/inventories.spec.ts tablist assertions, tests/items-crud.spec.ts Add-Items dialog tabs).

1. **No `data-state` on tabs.** The pinned contract is `data-state="active"/"inactive"` on each `role=tab` (specs assert it; the port's CSS keys on it). ui.Tabs emits `aria-selected` + `data-fui-tab-index` only. Mirroring data-state alongside aria-selected (or a StateAttrs opt) would unblock.

2. **No `aria-controls` linking.** Tabs and panels aren't wired by `aria-controls`/`id` pairs — AT parity with the rest of the port's tablists can't be expressed.

3. **Panels stay in the DOM (CSS-hidden), no content vacating.** The port's tabswitcher vacates hidden panel innerHTML (stashing it for re-show) so text locators scoped to the page don't match hidden panels — several suites depend on visible-position indexing. A VacateHidden option (or runtime behavior) that moves hidden panel content out of the DOM is needed for locator-sensitive hosts.

Suggested shape: `TabsConfig{StateAttrs bool, VacateHidden bool}` plus per-tab ID wiring, mirroring how Sidebar grew GroupMarkup/Open.

The port (go-collect-fastr) is the acceptance test.

DonaldMurillo:
Closes #320. Three opt-in knobs on `TabsConfig`, each zero-valued to today's
output. `ui.Menu` untouched — #319 is a separate, structural change.

## What ships

- **`StateAttrs`** — `data-state="active"/"inactive"` on every `role=tab`, the
  contract Radix-style ports pin locators to. Core already mirrors
  `aria-selected` on the same write; the module keeps `data-state` in step.
- **`ID`** — `id` + `aria-controls` wiring the tab/panel pair. This is AT
  parity the component could not express at all.
- **`VacateHidden`** — hidden panels ship empty, content parked in an adjacent
  stash script, so page-scoped locators cannot match text inside a hidden
  panel.

## The zero value is the load-bearing constraint

Pinned against the **exact pre-change bytes**, captured from `main` before any
edit — not by substring. Forcing either `StateAttrs` branch on fails it:

```
TestTabsZeroValueOutputPinned: zero-value Tabs output changed
```

Both branches mutation-proved independently. All 16 pre-existing Tabs tests
pass unchanged.

## Item 3 did not move the budget line

Vacating is runtime behaviour, and the core bundle had **20 gzipped bytes** of
headroom (12780/12800 gz6, 14766/14784 gz1). A kernel scanner-table entry does
not fit in that.

Raising the budget to make a feature fit is the exact failure mode that test
exists to prevent, so the module rides the existing `data-fui-prefetch` bridge
instead — whose documented purpose is "module ready by the time the user
clicks", and both behaviours here are interaction-time. **Core delta: zero
bytes**; `runtime.js` and `frag/` are untouched, verified by `git status`. The
new module is 681 gz6 against a 3072 per-module budget, fetched only on pages
that opt in.

## The caveat that buys, stated where a caller will see it

The module loads on first hover/focus, so a **programmatic** switch (SSE or
poll-driven signal write) on a strip nobody has touched shows an empty panel
until the first interaction heals it. SSR is unaffected — the initially-active
panel always ships with content — so this reaches only apps that move tabs
without the user.

I moved that onto the `VacateHidden` field doc rather than leaving it in the
architecture notes. Someone reading the field to decide whether to switch it on
should not have to find the caveat elsewhere.

Restoring **moves live nodes** rather than re-rendering the stash, so island
content and form state survive re-show with node identity intact — proven by an
e2e mutation where re-rendering from the stash loses a swapped island and typed
input.

## Hard rule 5

Three new `data-fui-*` attributes (`tabs-state`, `tabs-vacate`, `tabs-stash`) →
`core-ui/ARCHITECTURE.md` rows, `fragments.go` ownership, and the runtime test
suite. Docs updated in `runtime-contract.md`, `interactive-patterns.md`,
`ui-new-components.md`.

## One honest note from the mutation pass

The first attempt at the stash-escaping mutation **passed**: dropping only the
`</` → `<\/` rewrite changes nothing, because `json.Marshal`'s default
`<` escaping is the load-bearing mechanism and the rewrite is
carousel-consistent defence in depth. The honest mutation — no HTML escaping at
all — fails with the early-`</script>`-termination failure it exists to
prevent.

## Verified

`go build`, `go vet`, `gofmt`, `make analyze` clean. `framework/ui` green;
`core-ui/...` 35 packages green (547s) including the budget test and both new
e2e tests. Chromedp example suites not run.


DonaldMurillo:
## Independent review: HIGH bug in the wiring path, and the two things I flagged as risky were both clean

The review went straight at the two places I said a real bug would hide — and found neither. The stash escaping is sound (adversarial payloads including `</script>`, `</ScrIpT >`, `<!--` and pre-escaped `<` all round-trip byte-exact, zero literal `<` in the stash body), and the zero-value pin is genuine: it regenerated `ui.Tabs` from an `origin/main` archive and got byte-identical output, so the pin is not self-satisfying.

**The bug was in the path nobody tested.**

### HIGH — the scanner never matches the scope root

`tabs.js:84` uses `root.querySelectorAll(sel)`, and core's DOM-insertion observer hands each module scanner the **added node**, not a container. So a strip wrapper that *is* the added node is invisible to the scan. Demonstrated twice in chromedp:

- **Nested strips** — an outer vacate panel's restore sets `innerHTML`, the inner strip is the added node, never wired. Clicking its tab flips `data-active`, core mirrors `aria-selected`, CSS reveals the panel — and the panel is **permanently empty**, content inert in the stash. It looks alive with no body, and only heals on the next SPA navigation.
- **An island/RPC region swap** whose response root is a strip: identical.

What makes this a fair hit rather than bad luck: **the repo already pins this exact contract.** `module_root_marker_e2e_test.go` says a scanner "must match the node ITSELF", and `poll.js`, `computed.js`, `infinitescroll.js`, `animate.js` and `reveal.js` all use the stronger idiom. `tabs.js` followed the weaker one, and nothing covered post-load insertion, so nothing objected.

### It also resolved my own reported no-op honestly

I flagged that dropping the `</` → `<\/` rewrite passes. The review proved *why*: `json.Marshal` cannot emit `</` at all, so the `ReplaceAll` never finds input — it is a provable no-op as shipped, defence in depth only against a future switch to `SetEscapeHTML(false)`, which the existing test pair would catch. That is a better answer than the one I gave.

### And it corrected my own doc wording

Two ways. "Dropped until re-show" reads as suspend-and-replay; in fact in-flight updates are dropped **permanently** and re-show resurrects stale detached nodes. And the caveat I added to the field doc says "SSE or poll-driven" when the real class is broader — **any** hydration-time signal value differing from SSR writes `data-active` before interaction. It also checked the cases I had *not* claimed and confirmed deep links, anchors, restored scroll and autofocus are genuinely unaffected, since SSR always populates the active panel.

### Two more vacuity gaps

Dropping `scanAndLoadCSS(p)` passes everything; so does dropping the initial `apply(w)` — which is the exact line that exists for the programmatic-switch path. Both getting coverage.

All of the above is in flight, plus the CHANGELOG entry, the now-false `fragments.go` header comment, and a decision (not silence) on the failed-fetch case: `tabs` hangs *content visibility* on the prefetch bridge, and unlike `rpc` it fails silently.

DonaldMurillo:
## All findings closed — and a correction to this PR's own headline claim

`d9638c55`. The HIGH is fixed with the `poll.js` idiom and a nested-strip e2e; reverting the root-match line fails it (`inner strip nested in a restored vacate panel must be wired … timeout`), verified independently. Both vacuity gaps now have coverage, and both previously-passing mutations fail.

**This PR is no longer zero core bytes.** It was, before this commit. The prefetch retry costs **+5 gzipped bytes**, so the body's "Core delta: zero bytes" now describes the previous commit rather than the branch. Correcting it here rather than leaving a claim that was true when written and quietly stopped being so.

The budget **line** still does not move: 13 bytes of real headroom remain, and a stale comment claiming 33 (measured when the bundle was 14751; it is 14771) is corrected in place. That comment mattered — real headroom was tighter than it advertised, and two earlier formulations of the fix broke the cliff fixture before one fit.

### On finding 6, I asked for a decision and got the better one

The reviewer offered fix-or-justify. It fixed: the bridge marked an element attempted *before* knowing whether the fetch succeeded, so one failed request disabled the feature for the page lifetime with no signal. The reasoning is what makes it right — every marker-driven module self-heals on the next scan, and `tabs` is the only marker-less one, so the bridge was its single loader with a one-shot fuse. **That asymmetry was the bug**, not the failure itself. It benefits every prefetch consumer, not just tabs.

### Three corrections to my own prose

- "Dropped until re-show" implied suspend-and-replay in **five** files, not the three the review named. Nothing queues: re-show resurrects the pre-vacate nodes and only later updates land.
- My programmatic-switch caveat said "SSE or poll-driven"; the class is any signal write before module load, hydration included.
- It confirmed the cases I had *not* claimed — deep links, anchors, restored scroll, autofocus are genuinely unaffected, since SSR always populates the active panel.

That is three prose claims in this PR alone that outlived the code they described, all found by someone re-deriving them rather than reading them.

DonaldMurillo:
Independent review of the fix commit (`d9638c55`) came back with one blocking item and two I am declining. Dispositions:

**Accepted — stale twin comment.** `framework/ui/interactive_components.go:226` still said "~20 bytes" after the same claim was corrected in `tabs.js` to 13. Both cite the same constant. Fixed in `d4880903`. The measured figure at this head is 14771 against the level-1 line at 14784.

**Declined — multi-name partial-failure pin.** `_prefetch` marks the node attempted when *any* listed name resolves, so `data-fui-prefetch="a b"` with `a` failing pins `a` for the element's lifetime. Real, and not a regression: the old code retried nothing at all. Declining because nothing emits a multi-name value — tabs is the only in-tree producer and it always emits the single name `"tabs"` — and the fix costs core bytes out of 13 remaining. Spending the budget on a shape with no producer is the wrong trade. This belongs in the same change as the first multi-name emitter, not before it.

**Declined — failed `<script>` elements accumulate.** `loadModule` never removes the element on error, so each failed hover leaves one in `<head>`. Bounded by hover count, no network cost, and cleanup is bytes in the same budget. Same reasoning.

The parts I most wanted ruled out were ruled out by measurement rather than argument, so recording them here:

- **No request storm from the retry change.** In-flight re-hovers dedupe on `_modulePromises`; after a settled failure it is one fetch per pointer re-entry, which is the retry-per-event semantics the marker scanner and interaction bridge already ship.
- **`runtime.js` regeneration proven by construction.** `TestComposedRuntimeMatchesOnDiskFile` passes at both the parent and this head, and `boot.js` is the only changed fragment — so the `runtime.js` delta is exactly the `boot.js` delta. A hand-edit fails the head guard; a stale regen fails one of the two.
- **The programmatic-switch test's worry is disproven by assertion, not assumption.** It asserts `loadedModules.tabs === false` after the synthetic click, so a click that happened to arm the bridge would fail the test rather than quietly making it vacuous.
- **The old budget comment was wrong in both directions at once** — its prose said "clearing by 39" while its own trailing note said "33 over". The parent's true value was 14766. The correction was earned.

Each of the three new e2e tests was killed by its own mutation: removing the root-match line, restoring eager-marking, and dropping the initial `apply(w)`.

PR #328, opened 2026-08-30
ui.MenuItem: an addressable id

### LINKED ISSUE #327: ui.Menu: disabled rows emit tabindex="-1"aria-disabled="true" with no separating space
## What

`framework/ui/menu.go:218-222` builds the disabled attributes without a
leading space:

```go
disabledAttr := ""
if it.Disabled {
    disabledAttr = `aria-disabled="true"`
    ...
}
```

and line 249 concatenates it directly after the tabindex value's closing
quote:

```go
` role="menuitem" tabindex="` + tabindex + `"` + disabledAttr + rpcAttr + extra + `>`
```

So every disabled menu row renders:

```html
tabindex="-1"aria-disabled="true"
```

with no separator between the two attributes. Every other attribute in that
builder carries its own leading space; this one does not.

## Why it matters, and why it hasn't bitten

Browsers' error recovery splits the attributes anyway, so disabled menu items
work in practice — which is why this has survived. It is still invalid HTML,
and it breaks things that parse rather than recover:

- strict/XML-ish parsers and some HTML linters,
- the repo's own contracts pipeline if it ever grows a well-formedness rule
  (`core-ui/registry.WrapHTML` already rejects unterminated open tags — this is
  the same family of defect, on the attribute side rather than the tag side),
- anything diffing rendered markup for equality.

## Scope

Only the disabled path in `writeMenuItem`. `rpcAttr` and `extra` in the same
expression both carry leading spaces, so they are fine.

Worth grepping for the shape while fixing: an attribute string built without a
leading space and concatenated after a quote is a pattern that could exist
elsewhere in the component set. The invariant is "every attribute fragment owns
its leading space", and it is checkable.

## Not fixed in #327

Found while adding `MenuItem.ID` there. Deliberately left alone: the fix
changes zero-value rendered output, and that PR's entire safety argument is
that zero-value output is byte-identical to `main`. Correcting the spacing in
the same change would have meant the byte-pin could no longer be captured from
`main` — so the pin now records the malformed bytes with a `NOTE:` naming this
issue, and should be updated in the same commit that fixes it.

## Provenance

Surfaced by golden-byte capture during #319 item 3. Not a regression — `main`
emits it today.

DonaldMurillo:
Item 3 of 3 on #319, shipped alone. Submenus and `menuitemradio` are structural changes to `MenuItem` and are sequenced separately — this unblocks the port's id-pinned rows now, for a fraction of the work.

## What

`MenuItem.ExtraAttrs` drops `id` by design, so a row a host addresses by id — a Help Mode toggle whose page JS binds to it, a row that opens an activity sheet — could not survive a migration onto `ui.Menu`. An explicit `ID` field gives that one owner.

Three decisions, all following precedent already in the file rather than inventing:

- **`id` stays dropped in `ExtraAttrs`.** `SafeExtraAttrs` already removes every case variant unconditionally, so the drop needs no new code and cannot regress per component. The field doc now points at `ID` instead of implying ids are unsupported.
- **Uniqueness is the caller's**, stated rather than enforced — the component sees one menu, not the page, so document-level uniqueness isn't checkable here. Same stance `Class` already takes.
- **An id on a separator is ignored**, matching how this struct treats `Label` on a separator and `Confirm` on a non-RPC item. Panics here are reserved for structural emptiness (empty `Items`, missing `Label`).

## The golden bytes came from main, not from this branch

The zero-value pin is the whole safety argument for existing callers, so its provenance matters: captured **before any edit**, through a throwaway module importing the clean worktree so the repo tree stayed untouched, with determinism checked by two identical runs first. A pin taken after editing would be self-satisfying and indistinguishable from a correct one.

Five mutations, each watched failing and restored — including the case-variant one: removing `id` from `SafeExtraAttrs` renders `id="real-id" ... ID="smuggled2" Id="smuggled3"`, the duplicate-attribute fold-back this guard exists to stop.

## Filed, not fixed: #327

Golden-byte capture surfaced a pre-existing defect — every disabled row emits `tabindex="-1"aria-disabled="true"` with no separating space. Invalid HTML that browsers recover from, which is why it survived.

Deliberately left alone here: fixing it changes zero-value output, and this PR's safety argument is that zero-value output is byte-identical to `main`. The pin records the malformed bytes with a `NOTE:` naming #327, to be updated in the commit that fixes it.

## Verified

`go build`, `go vet`, `gofmt`, `make analyze` clean. `framework/ui` and all of `core-ui/...` green. `ARCHITECTURE.md` deliberately untouched — a plain HTML id is not a new `data-fui-*` attribute, so hard rule 5 doesn't fire.

## Summary by CodeRabbit

* **New Features**
  * Added support for assigning an addressable HTML `id` to menu rows through `MenuItem.ID`.
  * IDs are safely escaped and work with buttons, links, and RPC menu items.
  * Empty IDs preserve existing output; separator items ignore IDs.

* **Documentation**
  * Updated UI component documentation and the unreleased changelog with `MenuItem.ID` behavior and usage details.

* **Tests**
  * Added coverage for ID rendering, escaping, separators, and backward-compatible output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,13 @@ stabilises). Breaking changes are clearly marked with **BREAKING**.
   supported way to attach attributes to a component — cannot reach it.
   Inert for everyone else.
 
+- **`ui.MenuItem.ID`** sets the rendered menu row's `id` attribute, so page JS,
+  test suites, and aria wiring elsewhere on the page can address one exact row
+  (a Help Mode toggle a script binds to, an Imports row a shortcut targets).
+  Uniqueness is caller-owned like any HTML id, an empty value emits no `id`
+  (output unchanged), separators ignore it, and `MenuItem.ExtraAttrs` still
+  drops `id` — the field is the single owner.
+
 ## [0.76.0] - 2026-08-30
 
 ### Added
--- a/framework/docs/content/ui-new-components.md
+++ b/framework/docs/content/ui-new-components.md
@@ -117,7 +117,7 @@ raw (enumerated in `framework/ui/extraattrs_contract_test.go`). See
 - **breadcrumbs**: `core-ui/patterns/breadcrumbs`, `<nav aria-label=Breadcrumb>` trail
 - **pagination**: `core-ui/patterns/pagination`, numeric page navigation
 - **sidebar**: `framework/ui.Sidebar`, responsive primary nav with persistent, collapsible (local-storage persisted), off-canvas, and auto-hide variants; `Collapse` moves the collapsed state to the server, `GroupMarkup` swaps `<details>` groups for `button[aria-expanded][aria-controls]` + `hidden` container, `CollapseLabel`/`ExpandLabel` rename the toggle (see [Sidebar: server-owned collapse state](#sidebar-server-owned-collapse-state)); set `NavLabel` when a page has multiple navigation landmarks and mount the matching drawer with `MountSidebar`
-- **menu**: `framework/ui.Menu`, keyboard-driven dropdown built on `<details>`
+- **menu**: `framework/ui.Menu`, keyboard-driven dropdown built on `<details>`; `MenuItem.ID` gives a row an addressable `id` (caller-owned uniqueness, ignored on separators; `ExtraAttrs` still cannot set `id`)
 - **tabs**: `core-ui/patterns/tabs`, `<details>`-based tab strip, zero JS
 - **tree**: `core-ui/patterns/tree`, recursive tree with roving tabindex + lazy-load
 - **toc**: `framework/ui.TableOfContents`, auto-built sticky nav from `<h2>` / `<h3>`
--- a/framework/ui/menu.go
+++ b/framework/ui/menu.go
@@ -61,15 +61,25 @@ type MenuItem struct {
 	// Label and other fields are ignored when true.
 	Separator bool
 
+	// ID becomes the rendered row's id attribute, so page JS, test
+	// suites, or aria wiring elsewhere on the page can address this
+	// exact item (a Help Mode toggle a script binds to, an Imports
+	// row a shortcut targets). Uniqueness is caller-owned, like any
+	// HTML id: duplicates are the caller's bug, not enforced here.
+	// Ignored on separators, like every other field. Empty emits no
+	// id, leaving the output identical to a menu that never set it.
+	ID string
+
 	// Class appends to the rendered item's class list (rare; mainly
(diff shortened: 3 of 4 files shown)


DonaldMurillo:
@coderabbitai review

Small and self-contained, but two things worth attacking: (1) the byte-identical zero-value pin — its golden bytes were captured from `main` before any edit via a throwaway module, so the question is whether the capture really covers every `writeMenuItem` branch or leaves one unpinned; (2) whether `id` genuinely has one owner now — an `id` smuggled through `ExtraAttrs` in any case variant must still be dropped, or a row renders two.

DonaldMurillo:
The automated reviewer skipped this PR — zero reviews, zero line comments, and a "skip review" marker. The review gate reports "0 unresolved" identically for *reviewed clean* and *never reviewed*, so this looked reviewed and was not. Put an independent reviewer on it instead. It came back **merge**, no findings, with the two things I doubted settled by execution:

**The byte-identical pin is genuine, not self-satisfying.** This was the real question: golden bytes captured *after* an edit agree with the edit and are indistinguishable from a correct pin. The reviewer checked out `e7deb99b` (pre-`ID` main) and replayed the whole `goldenMenus` table there — all 10 subtests pass byte-for-byte, so the goldens provably predate the change. Every `writeMenuItem` branch is reached by the pin: separator, plain button, anchor, refused-scheme anchor, danger+disabled+icon, disabled anchor, RPC with method+confirm, RPC default, href+confirm-inert, item ExtraAttrs. The one unpinned input is `MenuTopStart`, already covered by `TestMenuPositionClasses`.

**`id` has exactly one owner, across every case variant.** `SafeExtraAttrs` drops `id` via a built-in denylist, with `isSafeAttrKey` as a second gate. Verified by rendering `{"id","ID","Id","iD","i d","İD"}` through ExtraAttrs onto button, anchor, RPC, and disabled rows plus a separator carrying both — zero survivors, one `id="` in the output and it belongs to the panel div. The separator returns before either field is consulted.

**Vacuity, four mutations run:** deleting the `idAttr` emission entirely fails three tests; zero-value emitting `id=""` fails the golden; dropping `render.Escape` fails the escaping test; reordering `id` after `role=` fails the emission test. The feature cannot be deleted with the suite green.

One thing worth recording: the missing space in `tabindex="-1"aria-disabled="true"` is **pre-existing on main** — the reviewer's main replay proves it — and this PR pins it as-is with a comment rather than fixing it inline. That is the right call for scope, and the golden now doubles as a tripwire for #327, which owns the fix.

Issue #331, opened 2026-08-31
framework/ui's init breaks TestComponentCSS_* in any package that imports it


DonaldMurillo:
## What

Importing `framework/ui` from a test in `framework/uihost` breaks
`TestComponentCSS_*` in that same package, without either test file referring to
the other.

`framework/ui`'s package `init` registers `ui-button`, `ui-page-header`, and
`ui-sidebar` as `LoadAlways` in the **process-global** style registry. The
`TestComponentCSS_*` tests assert a single-component premise about what that
registry contains, so the mere presence of the import in any file in the package
flips them red.

## How it was found

While adding the internal-link strict check (#313). The test fixture there wants
to reproduce the blueprint generator's real `ui.SiteFooter` markup, which means
importing `framework/ui`. Doing so turned `TestComponentCSS_*` red.

Confirmed it was the import and not the change under test with a three-way
isolation run:

| what was present | package suite |
|---|---|
| pristine tree | pass |
| production diff only | pass |
| production diff + the test file importing `framework/ui` | **fail** |

## Why it matters

The coupling is invisible at the point of use. Nothing in `framework/uihost`
says "do not import `framework/ui` here", the failure surfaces in an unrelated
test, and the cause is a package `init` two hops away. The next person to reach
for a realistic UI fixture in this package pays the same debugging cost.

It also has a real cost already: the #313 fixture is hand-written anchor markup
approximating `ui.SiteFooter` rather than the component itself, so nothing
enforces that the two agree. That is stated in a comment there rather than
tested, which is the weaker thing.

## Fix shapes

1. **Make `TestComponentCSS_*` state its premise instead of assuming it** —
   snapshot the registry and assert on the delta, so registrations made by any
   imported package are not mistaken for the component under test. Narrowest
   fix, and it makes the tests honest about a global they do not own.
2. **Stop registering `LoadAlways` styles from `init`** and move it behind an
   explicit call. Correct in principle, wider blast radius, and `LoadAlways` may
   exist precisely so callers do not have to opt in.
3. **Leave it and document the constraint** in the package. Cheapest, and it
   keeps the trap.

(1) looks right: a test that breaks when an unrelated package is imported is
asserting something it did not mean to assert.

## Provenance

Found by an implementation run for #313, verified by isolation rather than
reasoning. Pre-existing on `main` — no change caused it.

PR #332, opened 2026-08-31
strict: flag chrome links nothing serves; fix malformed menu attribute

### LINKED ISSUE #313: Launch checks: extend WithStrict instead of adding a second gate for public-app essentials
The idea: an app that is about to face the public should not build green while it is missing the pages and metadata a public app needs. Sign-in, sign-up, forgot-password, terms, privacy, about, contact; sitemap, robots.txt, per-page titles and JSON-LD; reduced motion and the rest of the accessibility bar. Opt out explicitly, or the build fails.

## Most of this already ships

`uihost.WithStrict` (`framework/uihost/strict.go`) enforces at Mount: screen titles, screen descriptions, site description, site icon, sitemap, robots.txt, and axe coverage per route. Per-check `enforce`/`warn`/`off`, `ExemptScreens` route patterns, and a panic listing every finding at once with fix hints. Its doc comment already calls this "launch hygiene".

`framework/contracts` + `gofastr verify` owns the static rules, with a suppression grammar (`//gofastr:allow(ID) <reason>`) and meta-rules `GOFASTR0001`–`0004` that make a reason-less, stale, typo'd, or blanket suppression an error in its own right.

So this is a posture change to `WithStrict` plus a few new checks, not a new subsystem. The thing to avoid is two systems reporting the same finding through two different escape hatches.

## The list splits three ways

**Verifiable, framework owns the mechanism.** Titles, descriptions, sitemap, robots, JSON-LD, accessibility. Done except JSON-LD: `core-ui/seo` renders it and `framework/uihost/organization.go` injects an Organization block sitewide, but nothing checks that a page declares any of its own.

**Conditional on what the app enabled.** `battery/auth` registers `POST /login` (`core.go:63`) and `POST /forgot-password` / `POST /reset-password` (`password_reset.go:103`–`104`), and ships no GET screens for them. The check worth writing is: the reset endpoints are mounted and no screen posts to them. It has to key on the auth battery being mounted with password credentials. An app with no auth being told it needs a login page is the false positive that gets the whole gate switched off.

**Content the framework cannot judge.** Terms, privacy, about, contact. An existence check here is vacuous — a screen returning `TODO` passes forever, and the build goes green over a legally worthless page. Privacy policies are also jurisdiction-specific.

## What to check instead

**Internal link integrity.** An href in the app's own chrome or screens pointing at a path no route registers. Answerable from the route table, not satisfiable with a stub, and it catches a much wider class than legal pages. #312 is a live instance: the blueprint generator emits footer links to `/terms` and `/privacy` and never registers those screens.

**Scaffold, then check the scaffold.** Have the blueprint generate `/terms`, `/privacy`, `/about`, `/contact` with real structure and a placeholder marker, and add a contracts rule that fires while the marker survives. "You shipped the placeholder" can fail. "Does the route exist" cannot.

## Open questions

1. **Severity default.** Erroring the build by default works against first-contact cost: a hello-world that will not compile until there is a privacy policy is a bounce, and the top search result becomes how to turn it off. Suggested default is warn, with `WithStrict` staying the switch that makes them errors — it already means exactly that.

2. **Where it runs.** `contracts` is static; `WithStrict` needs a mounted app to see the route table. Failing the *build* means either teaching contracts to read screen registrations from the AST, or having `gofastr build` boot the app in check mode. The v0.66 boot gate is precedent for the second and is the cheaper path.

3. **Reduced motion.** `core-ui` already respects `prefers-reduced-motion` in patterns/tabs, accordion, skeleton, progress, infinitescroll, and the runtime signal flash. Apps ship zero CSS under hard rule 7, so a per-app check duplicates `GOFASTR1801` (bespoke CSS) from the far end. If there is a gap it is suppressed bespoke CSS that animates, which is a narrowing on the existing rule.

4. **Escape hatch.** Reuse the contracts suppression grammar and per-rule severity in `gofastr.contracts.yml`. A second opt-out vocabulary would split `gofastr verify --explain` across two catalogs.

### LINKED ISSUE #327: ui.Menu: disabled rows emit tabindex="-1"aria-disabled="true" with no separating space

DonaldMurillo:
Two tickets, grouped to share a CI run.

Closes #313. Closes #327.

## #313 — internal link integrity in strict mode

Most of the original launch-checks list should not be built, and the reasoning is on the issue. Existence checks on terms/privacy/about/contact are vacuous — a screen returning `TODO` passes forever and the build goes green over a legally worthless page. A reduced-motion check duplicates `GOFASTR1801` from the far end, since apps ship zero CSS under hard rule 7.

What survives is the one item answerable from the route table and not satisfiable with a stub: **a route reference the app emits that points at a path nothing serves.** #312 was the live instance — footer links to `/terms` and `/privacy`, neither registered.

Default posture is unchanged: this is a `WithStrict` check, so it only runs for apps that opted in, and it is tunable through `StrictConfig.InternalLinks` like every other check. No second opt-out vocabulary.

**Rendered screen output is deliberately out of scope**, with evidence rather than a shrug:

- `Router.RenderRaw` skips `Load` (`core-ui/app/router.go:307-334`), so a `ScreenLoader` screen rendered without a request is not the page users get — its links can differ.
- Real rendering goes through `App.RenderPageResult` → `Load(ctx)`, a request-context data hook touching auth and the DB. At Mount there is no request and possibly no database.
- The codebase's own headless render-everything path, `UIHost.RenderStaticPage`, needs a live ctx and still skips policy-blocked screens.

A screen-rendering boot gate would either demand a database at boot or silently under-cover. Chrome alone is the honest deliverable.

**The exclusion set is what decides whether this feature survives contact** — one false positive gets the whole gate switched off. Each has a test: external URLs, any scheme (`mailto:`, `tel:`), protocol-relative, backslashes, pure fragments, query-only, fragment/query stripped before resolution, template placeholders (`{slug}`, `:id`, `:path*`, `<year>`), relative references, the host's own `/__gofastr/` prefix, static files, artifact endpoints gated exactly as Mount gates them (so an unconfigured `/sitemap.xml` link stays a finding), framework-router GET routes, and dynamic plus catch-all routes through `Router.Resolve`.

Sixteen mutations run, each observed red.

### One thing I corrected in review

The vacuity fixture is hand-written anchor markup approximating `ui.SiteFooter`, and its comment claimed it mirrored the real component "down to the anchor markup it emits". Nothing enforced that. The comment now says what is true: it pins the *check*, not the component, and fidelity to the generator's real output is pinned where `framework/ui` is importable — the marketing-link test in `cmd/gofastr`.

It is hand-written for a reason worth knowing about, now filed as #331: importing `framework/ui` into a `framework/uihost` test breaks `TestComponentCSS_*` in the same package, because `framework/ui`'s `init` registers three components as `LoadAlways` in a process-global registry. Proven by a three-way isolation run, pre-existing on main.

## #327 — malformed attribute on disabled menu rows

`aria-disabled` was concatenated straight onto `tabindex="-1"` with no separating space, so a disabled row rendered `tabindex="-1"aria-disabled="true"`. Every other attribute chunk in that builder carries its own leading space; this one did not.

The byte-identical goldens added by #328 pinned the malformed form on purpose, as a tripwire for this fix. It worked — removing the space again reddens `danger-disabled-icon-class`, `disabled-anchor`, and `TestMenuItemIDEmittedOnRows`. Watched both directions.


DonaldMurillo:
Independent review came back **merge-with-fixes** with a High I'm glad it went looking for. Posting the dispositions before the fix lands.

## Accepted — the check runs too early to be right

**F1 (High).** The check resolves chrome links at the top of `Mount`, and the route table is not complete there. Batteries initialise during `App.Start` (`framework/app.go:2300-2303`), and the admin battery registers `/admin`, `/admin/queue`, `/admin/audit` from `Init`. So does the documented mount order for `sdkdocs.Mount`, `static.Mount`, `widget.Mount`.

Executed, not argued: a footer link to `/sdk/docs` with the route registered in the documented order **panics strict boot**, while the same app without `WithStrict` serves it 200.

The scenario that makes this fatal: a `gofastr generate` app ships strict ON, registers the admin battery, and has a sidebar "Back office" → `/admin`. Working link, panicking boot. And the remedy a user reaches for — `ExemptScreens: ["/admin", "/admin/*"]` — exempts exactly the links most likely to rot, which defeats the check.

I asked the reviewer for a legitimate app shape this wrongly flags, precisely because one false positive gets the whole gate switched off. This is that shape, and it is the default generated app.

**The fix is to defer, not to document.** A check whose entire value is "answerable from the route table" is wrong when the table is partial, and no amount of prose fixes a gate that fails working apps. Only the link check moves; the other seven checks do not need the route table and stay at Mount.

**F2 (Medium).** The comment claiming `/__gofastr/` covers UIHost's own endpoints is false — it misses `/llm-pages.md`, `/llms-full.txt`, `/.well-known/agent-card.json`, `/.well-known/agent.json`, and `/.well-known/jwks.json`. Each verified: panics with strict, serves 200 without. Deferring the check fixes this for free, since those register inside `Mount`; the comment gets corrected either way.

**F3 (Low).** `href="/docs/caf%C3%A9"` against a screen at `/docs/café` is flagged, while the runtime serves it 200 — net/http decodes before routing and the check does not.

**F5 (Info).** A mutation the PR did not declare: dropping `stdhtml.UnescapeString` in `extractChromeHrefs` passes the whole suite. Entity-escaped hrefs are unpinned. Assertion added.

## Accepted as a wording fix, not a code fix

**F4.** A catch-all `/{path...}` makes the check vacuously pass — a footer link to `/definitely/missing` boots clean and 404s at runtime. Two claims overstate: `strict.go:467` ("a link a user can click that 404s is a finding") and the CHANGELOG's "any GET route on the framework router".

Not defeating catch-alls. A catch-all *may* serve the path, and guessing produces false positives — the failure mode this whole review was aimed at. The honest move is to say a link under a catch-all is not verified.

## Worth recording from the ruled-out list

The two things the PR declared as not mutation-tested were checked and are **true rather than convenient**: `contextComponentFunc.Render()` delegates with `context.Background()`, so `SafeRender` and `SafeRenderCtx(background)` are observationally identical at boot, and the nil-slot removal only trades a silent skip for a recovered-panic warn. Neither is killable by behaviour.

Also confirmed as correct flags, not false positives: trailing-slash mismatch, case mismatch, and an unconfigured `/sitemap.xml` — each 404s at runtime.

For #327, the reviewer swept the bug class rather than the instance: 32 rendered configurations across every hand-built builder in `framework/ui`, regex-scanned for attribute-adjacency defects. No other instance exists.

PR #336, opened 2026-08-31
ui.Menu: submenus and menuitemradio rows

### LINKED ISSUE #319: ui.Menu cannot express submenu + menuitemradio contracts (nested role=menu, aria-haspopup items, id-addressable rows)

DonaldMurillo:
Closes #319 — the last open ticket on the board.

`MenuItem` gained `Children`, `Radio`, and `Checked`. A parent with `Children` discloses a nested `role="menu"` through the same `data-fui-disclosure` machinery the top level already uses, rather than a second mechanism.

## The keyboard contract ships with the markup

ArrowRight opens a submenu and moves focus into it; ArrowLeft closes it and returns focus to the parent row (swapped under RTL); roving focus and type-ahead stay scoped to the item's own panel; Escape closes one level at a time; Tab closes the whole chain.

This was a decision, not a default. The issue asked only for the markup, and I asked the porting side whether they needed keyboard navigation and got no answer. Deciding rather than blocking: `ui.Menu` navigates by keyboard today, so growing submenus without a keyboard path would make the component **worse for keyboard users than it currently is**. Markup-only was never a smaller version of this change; it was a regression wearing a feature's clothes.

## Illegal shapes panic instead of rendering something incoherent

`Children` alongside `Href`, `RPC`, or `Radio` panics at render. A submenu parent is purely a disclosure — a row that both opens a panel and navigates has no coherent behaviour, so it fails loudly at the point the developer wrote it.

A disabled parent renders inert with its children unreachable.

## What the work surfaced

The disclosure module's focus-on-open had to start skipping `<summary>` rows: a nested parent row matched the focus-first selector and stole focus back. That bug is only reachable once a `<summary>` can itself be a menu row, so nothing could have found it before this change.

## Zero-value output is unchanged

The ten existing goldens are untouched and passing, including `ExtraAttrs` id-dropping at depth. `positionsForHash` recurses only when `Children` exist, so auto-generated ids of childless menus do not move.

## Budget

Core runtime bytes unchanged — this is module work. The menu module grew 460 gzipped bytes and disclosure 51, both inside their budgets. Worth stating explicitly because core had only 8 bytes of level-1 headroom when this started, and a keyboard contract in core would not have fit.

## Mutations

Sixteen, each watched red and restored. The ones worth naming:

| mutation | caught by |
|---|---|
| kill `closest('[role=menu]') === panel` scoping | ArrowDown from a submenu row leaked to the top level |
| drop `menuitemradio` from the row selector | roving focus skipped radio rows |
| kill radio click arbitration | `aria-checked` map unchanged after click |
| kill the `aria-disabled` guard on ArrowRight | a disabled parent opened its submenu |
| revert Escape to close-all | first Escape closed the top level too |
| remove `:not(summary)` from focus-first | focus never landed in the panel |
| Tab closes only the innermost | outer menu stayed open |
| stray byte in plain-item output | the zero-value golden |

## Stated honestly, not run

- No RTL keyboard or `:dir(rtl)` browser test — the RTL branch is `getComputedStyle().direction`-driven and the caret degrades to the LTR glyph where `:dir()` is unsupported.
- No `examples/meridian` pass: no example page renders submenus or radios yet, so the visual check ran against a harness serving the component's real SSR and real registered CSS. Adding a demo to `examples/site`'s `/components/menu` page is worth a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Menus now support nested submenus with keyboard navigation, focus management, and RTL-aware controls.
  * Added radio-style menu items with checked states and automatic same-group selection updates.
  * Disabled submenu entries are skipped during navigation, while type-ahead searches visible labels.
* **Bug Fixes**
  * Improved nested disclosure closing with Escape and Tab, including correct focus restoration.
  * Fixed disabled-menu rendering and submenu sizing information.
  * Improved combobox filtering and CommandPalette closing and viewport behavior.
* **Documentation**
  * Updated menu accessibility, keyboard interaction, submenu, and radio-group behavior documentation.


DonaldMurillo:
All three actionable comments were already addressed in `48a3d4c`, which landed after the reviewed commit (`ccf8cadb`) — an independent review found the same three defects. Worth saying that plainly: **three-for-three convergence between two reviewers that could not see each other** is good evidence these were real rather than stylistic.

Verified each against the current head rather than assuming the commit covered them:

**1. `disclosure.js` focus-on-open — accepted, implemented as suggested.** The lookup is scoped to the disclosure's own panel via `closest('[role=menu]') === panel` (the same ownership rule `menu.js rows()` uses) and `:not(summary)` is gone, because a same-panel submenu parent is a legitimate first row.

The failure this fixes is worse than it reads: with the descendant search, a panel whose **first row is a submenu parent** put focus on a row inside the still-closed nested `<details>` — hidden, so `.focus()` was a silent no-op. The menu opened with `activeElement` on BODY and ArrowDown/Up/Home/End and type-ahead were all dead. A `menu_contract_e2e_test.go` fixture with a submenu as the first item now covers it, and fails if descendant searching returns.

**2. `menu.js` radio arbitration — accepted, resolved differently. Saying why.** You suggest dropping the `group === null` short-circuit so the comparison is purely `sib.getAttribute('data-fui-menu-radio') === group`.

That makes every group-less radio on the panel one implicit shared group, so activating one unchecks all the others. I went the other way — `group === null ? sib === r : ...` — so a group-less row is a group of one and self-checks.

The reason is that the comment beside the code already stated the intent ("implicit group of one, the row still self-checks") and the code contradicted it by unchecking every radio in the panel. One of the two had to move, and the comment's version is the safer semantic: silently coupling unrelated hand-authored rows because neither declared a group is a surprise, whereas a row that only checks itself is inert and obvious. `framework/ui` cannot produce a group-less radio at all, so this only affects hand-written HTML — where "no group" most likely means "not grouped", not "grouped with whatever else forgot".

Happy to flip it if you read the ARIA intent differently, but code and comment now agree either way, which they did not before.

**3. `menu_test.go` dead assertion — accepted, implemented as suggested.** The guard was inverted (`== false` on a string that always contains `aria-checked`), so the count never ran, and no fixture carried `Checked` without `Radio`. There is now a `{Label: "Ghost", Checked: true}` row, an assertion that it renders as a plain menuitem with nothing between `tabindex` and the label, and an unconditional `Count(out, "aria-checked") == 2`.

Proved dead before fixing: a mutation making `Checked`-without-`Radio` emit `aria-checked="true"` **passed** the old test and reddens the new one.

PR #341, opened 2026-08-31
Analyzers from the repo's bug history, and every failure they found

DonaldMurillo:
Mines the repo's own bug history — issues, PR review comments, past fixes —
for invariants worth enforcing mechanically, adds the analyzers that can
enforce them, and then fixes every failure the new gates and the red-test
sweep turned up.

## The analyzers

Six new type-aware passes in `internal/analyzers/`, registered in
`cmd/vettool` alongside ten stock `x/tools` passes:

| Analyzer | Invariant | Found |
|---|---|---|
| `unboundedbody` | an inbound `*http.Request` body is read or decoded with no size cap | 9 |
| `errleak` | internal error text on a 5xx response | 9 |
| `fieldtypeswitch` | a non-exhaustive `schema.FieldType` switch with no `default` | 6 |
| `hygiene` | empty error branches; client calls with no timeout | 0 (zero-backlog gates) |
| `reqparamlimit` | a request-sourced integer reaching a limit-shaped parameter unclamped | 0 |
| `discardmutator` | a security-state mutation whose error is dropped where the handler says success | 0 |

Two more are written and fixture-tested but **not** registered, with the
measurement that justifies it recorded in `cmd/vettool/main.go`:
`fmtformat` (4 findings, all false positives — the repo fixed the
encoded-pattern class at the consumer, where `ui.DataTable` and
`core-ui/patterns/pagination` substitute their `%s`/`%d` markers with
`strings.Replace` and never `fmt`, a posture the analyzer cannot see) and
`testgap` (40 findings, two of them on the analyzer sources in its own
directory — a test-quality heuristic, not an invariant).

All four of those arrived unregistered *and* unexplained: they passed
their own fixtures while `make analyze`, the pre-commit hook and CI's vet
step never ran them over a repo file.
`TestEveryAnalyzerIsWiredOrExplained` now fails on a package that is
neither registered nor named in a comment there.

Type information is what makes these worth writing rather than leaving to
the contracts pipeline: `unboundedbody` distinguishes `*http.Request` from
`*http.Response`, and grep could not — 28 candidates, 19 of them outbound.
`fieldtypeswitch` reads the constant set from the schema package at
analysis time, so a new field type breaks the switches that ignore it.

Two more were written, measured, and **dropped**, with the reasoning
recorded in `hygiene.go` so nobody re-derives it: `SprintfSQL` (287
findings, all legitimate identifier interpolation — placeholders bind
values, not identifiers, and contracts' GOFASTR1401 covers host apps) and
`HandlerContext` (6 findings, all correct code — an idempotency claim
released post-response, a cache write a disconnect must not abort, an SSE
stream outliving its request).

Most measured classes came back **zero**. The value here is pinning
discipline that already exists, not accumulating a backlog.

The existing gates were also leaking: `mapwriter` resolved package identity
by identifier name, so an aliased import walked past it, and it did not
know `maps.Keys`/`maps.Values` are map-ordered. `check-embed` stopped at a
component-returning helper instead of following it. `cmd/repolint` gained a
`crud-exposure-rederived` rule.

## The fixes

Every failing test in the tree, 109 of them. The changelog carries the
detail; the pattern worth naming here is that **six were the same rule
written twice and fixed once**:

- `kiln/db` kept its own `sqlDefault` carrying the injection arm
  `framework/migrate` deleted after a verified payload
- the styling-prop ban existed in `kiln/protocol` and `kiln/freeze`, and
  nowhere in journal replay
- the eval credential denylist was hand-copied into two runners and had
  drifted (`NUGET_`/`TWILIO_` in one only); neither caught a bare `*_TOKEN`
- the candidate-env allowlist sat one package away from the runner passing
  `os.Environ()` whole to agent-written code
- `router.Param`'s path sanitizer never reached `handler.Bind`
- `DBQueue` had no claim fence while its Redis twin did

Each is now one implementation with both callers pointing at it. That is
the durable half.

## Worth a careful look

**Tests changed, ~14.** Most were fixtures modelling something a real
client would not do. Four asserted the vulnerable behaviour directly: the
setup cookie test required `cookie == URL token` (the replay itself), the
log listing test required the ungated tools stay listed, and two pluginhost
tests pinned the CSP byte-for-byte. Three of those four were called for by
the red tests' own comments. Each is explained in its commit.

**Two red tests had real defects.** `raceClaimInterleave` returned its
re-claimed job by value before the hook that fills it ran, so every
assertion compared against a zero `Job`; it now fails loudly if the
interleave never fires. And two fmt-carry tests searched for `"q="+enc`
where `enc` already starts with `q=`, so no correct href could satisfy them.

**One budget line moved, by 2 bytes, and the margin is now a tripwire.**
The runtime module-name guard costs 27 gzip bytes at level 1 and the
bracket had 33, so the cliff self-test stopped crossing the congestion
window. The guard is written as cheaply as it goes — three other spellings
measured larger — and `budget_test.go` records what bought the bytes.

Re-measured on the merged tree: **level-6 is 12805 against a 12808 goal (3
bytes), level-1 is 14786 against a 14788 window (2 bytes).** That is not a
margin, it is a countdown. The next character added to `runtime.js` turns
CI red, and the fix the rule asks for is carving a core feature into a
demand module — a change to what loads on first paint, which is a design
call, not cleanup, so it is deliberately not made here.

**Guards proven, not argued.** The queue claim fence and the ambient-tx
update gate were each broken deliberately and watched fail before being put
back (CLAUDE.md rule 11).

## Verification

`./scripts/test-all.sh` green. `go vet ./...`, `make analyze`, `gofmt -l`
all clean.

Note for reviewers: plain `go test ./...` **cannot** confirm this tree.
`examples/site` takes ~1200s against Go's 600s default per-package timeout,
so it reports FAIL on time alone. `uie2e` and Meridian's axe sweep flake the
same way under full parallelism — `test-all.sh` caps `-p` for exactly that
reason.

---

Two notes on the branch itself, not the change:

- It carries one commit that is not part of this work: `d631c9a2`
  (`feat(ui): SidebarItem.Open`), which was sitting unpushed on local `main`
  when the branch was cut. Kept deliberately rather than rebased out.
- `origin/main` (through #344) is merged in and `test-all.sh` was green on
  the merged tree.
- **CodeRabbit has reviewed none of this.** Its `pass` in `gh pr checks`
  means it ran and skipped: 284 files against a 100-file cap, and the
  account is out of review credits. `pr-review-findings.sh --gate` reports
  "0 unresolved" for the same reason — there is nothing to resolve because
  nothing was reviewed. Treat the gate as vacuous here, not as clearance.
- CI had never run on this branch until 2026-08-31: the PR was
  `CONFLICTING` from the moment it was opened, so GitHub could not build a
  merge ref and never dispatched `pull_request`. The merge above fixed it.

https://claude.ai/code/session_01LP1Z771Td7PxCZxyRFnEVh


DonaldMurillo:
Flagging two things about this PR's review state, and one targeted check I did run.

**It has zero reviews and the gate reads green.** 13 checks pass and `pr-review-findings.sh --gate` reports "0 unresolved" — which it reports identically for *reviewed clean* and *never reviewed*. On a 19k-line change that alters request-body handling and 5xx response bodies, that green is not evidence.

I could not get an independent pass on it: `omp` is uninstalled on this machine (binary gone from every PATH dir, `~/.omp` state intact), so no worker is dispatchable right now. I am not going to imply a 19k-line diff got a review it did not get.

**What I did check**, because it is the one thing in here that can break a running app on upgrade: the body caps. A cap set too low on an upload path breaks every app that updates, silently and only for large payloads.

| cap | value | surface |
|---|---|---|
| `maxThemeEditBody` | 64 KiB | `cmd/gofastr` theme editor (dev tool) |
| `maxAgentBody` | 64 KiB | `cmd/kiln` agent HTTP |
| `maxWebSendBody` | 1 MiB | experimental harness web client |
| `maxPanelBody` | 1 MiB | kiln chat panel |
| `maxRESTBody` | 4 MiB | experimental harness **control** REST — command payloads, not CRUD |
| `maxBody` ×2 | — | analyzer testdata fixtures, not production |

**None of them lands on the CRUD or file-upload path**, which is where a wrong number would have hurt. `framework/file` is hardened separately in this PR (+128 in `filefield.go`, three new security test files) rather than capped by these constants. So the category I was most worried about is clean.

**What I did NOT review:** everything else. The nine `errleak` fixes (do any drop a diagnostic an operator needs?), the six `fieldtypeswitch` defaults (is each added default the right behaviour, or does one silently swallow a new field type?), the six analyzers themselves, and the ~19k lines of accompanying changes across `battery/auth` (35 files) and `battery/queue` (16).

Recording this so the gate's green is not mistaken for coverage. The PR is BEHIND main and will need a catch-up merge regardless.

DonaldMurillo:
Second category checked: the **errleak** fixes. Conclusion — no diagnostic is lost.

My concern was that removing `err.Error()` from a 5xx could take the cause with it, leaving an operator with a bare "internal error" and nothing to debug from. It does not. Every removal I found pairs with a `log.Printf` carrying the same error, so the cause moves from the client response to the server log — where operators read it and attackers do not.

Audited by walking each diff hunk and checking whether a hunk that removes `err.Error()` from a response also adds a log call:

| file | result |
|---|---|
| `core-ui/widget/server.go` | logged |
| `examples/blog/main.go` | logged |
| `examples/embed-demo/main.go` | logged |
| `framework/experimental/harness/control/mcpserver/http.go` | logged |
| `framework/experimental/harness/control/ws/ws.go` | logged |
| `framework/uihost/agentready.go` | logged |
| `kiln/render/render.go` | logged |

**7 hunks, 0 without a log.**

Two caveats, stated rather than glossed:

- The PR reports **9** errleak findings and my pattern matched **7** hunks. The other two are in a shape my regex missed — a JSON error body, a `fmt.Fprintf`, or a helper — so they are unchecked, not clean. Worth someone confirming those two specifically.
- `core-ui/widget/server.go`'s replacement text is a good example of the right instinct and worth keeping as the house pattern: *"The error names embed paths; the client gets the fact, the operator gets the cause."* That is the actual reason this class matters — the leaked text was filesystem structure.

Still outstanding on this PR: the six `fieldtypeswitch` defaults, the six analyzers themselves, and the bulk of the ~19k lines across `battery/auth` and `battery/queue`. Two categories down.

DonaldMurillo:
Third category: the **fieldtypeswitch** defaults. Shape is right; two notes.

The risk with adding a `default:` to satisfy an exhaustiveness analyzer is that it converts a *loud* gap into a *silent* one — the switch stops being flagged and starts quietly swallowing whatever it does not know. That is not what these do. Each one names what falls through and why that is safe:

| site | default |
|---|---|
| `framework/crud/crud_upload.go` | "Not a file-bearing field: nothing to collect" — inert |
| `framework/sdkdocs/screens.go` | falls through to the first-column fallback; only ordered types make a useful range-filter example |
| DataTable render path | escaped text via `render.Text`, **with an explicit warning** that a type whose stored value is a URL or path needs its own case, because this default has no scheme guard |
| `cmd/gofastr/blueprint.go` | routes to `seedConstraintRejection`, deliberately coercion-tolerant |

That third one is the pattern the others should be measured against. It does not just say "safe" — it names *what makes it safe* (escaped text) and *what would break it* (a scheme-bearing type). A future field type that renders a URL will hit that comment before it hits production.

Two caveats:

- The PR reports **6** findings; I located **4** switches with new defaults, one of which is the analyzer's own testdata. So 3 production defaults verified, and the remainder unchecked rather than clean.
- `cmd/gofastr/blueprint.go`'s is not a one-line default — it introduces `seedConstraintRejection` and re-routes validation through the boot validator instead of re-deriving constraints. That is a behavioural change wearing an exhaustiveness fix's clothes, and it deserves scrutiny on its own terms rather than as part of this category.

Three categories checked: body caps, errleak, fieldtypeswitch defaults. Still untouched: the six analyzers themselves, and the bulk of ~19k lines across `battery/auth` (35 files) and `battery/queue` (16).

DonaldMurillo:
Fourth category: **are the analyzers load-bearing, or do they pass because they cannot see?**

`make analyze` exits 0 on this branch, so all 24 findings really are fixed. But a clean run is exactly what a broken analyzer produces, and those two states are indistinguishable from the outside. So I reverted individual fixes and checked that the analyzer notices:

| analyzer | fix reverted | result |
|---|---|---|
| `errleak` | `ws.go` hijack error back into the 5xx body | **caught** (exit 2) |
| `unboundedbody` | `MaxBytesReader` removed from `theme_edit.go` | **caught** (exit 2) |

The unboundedbody message is the right shape too — it does not just flag, it says what to do:

```
cmd/gofastr/theme_edit.go:465:12: reads an inbound request body with no size cap:
  wrap it first (r.Body = http.MaxBytesReader(w, r.Body, max))
```

Restored, `make analyze` clean again. So these two gates would catch a regression rather than decorating one.

I did not revert-test the other four (`fieldtypeswitch`, `hygiene`, `reqparamlimit`, `discardmutator`). Worth noting that three of those shipped with **0 findings** — declared as zero-backlog gates — which is precisely the configuration where "works" and "sees nothing" cannot be told apart from a green run. Their fixtures presumably cover it, but fixtures test the analyzer against its own testdata, not against this repo. A reversion test on a real file is a different claim.

Also worth crediting: `TestEveryAnalyzerIsWiredOrExplained` failing on a package that is neither registered nor explained closes exactly the hole that let four analyzers sit unregistered *and* unexplained — passing their own fixtures while never running over a repo file. That is the same shape as a test whose fixture cannot express the failure, caught structurally rather than by noticing.

Four categories checked: body caps, errleak, fieldtypeswitch defaults, analyzer teeth. Remaining: the ~19k lines across `battery/auth` (35 files) and `battery/queue` (16), and the four un-reverted analyzers.

DonaldMurillo:
**Raising the priority of this PR.** `battery/auth` contains an account-takeover fix, and it is currently sitting in an unreviewed 19k-line change.

## What `token_purpose.go` fixes

Password reset, email verification and magic-link login share one `MagicLinkTokenStore` whose payload is an opaque string. Nothing recorded which flow minted a token, so **every handler redeemed any token the store held** and read the payload as whatever its own flow expected.

The sharp direction: a **verification** token — minted behind only the user's own session, 24h TTL, delivered as a low-sensitivity GET link — redeemed at `/auth/reset-password`, where its userID payload reached `SetPassword`. The reverse direction was its own bug: a reset token redeemed at magic-link/verify had its userID looked up as an email, missed, and **auto-created an account keyed by the victim's user id**.

## Why the implementation is worth the read

Two decisions in it are the kind that are wrong by default and right here:

**Peek before redeem.** `RedeemToken` is atomic and single-use, so checking the purpose *after* redemption would let anyone holding a victim's magic link **destroy it** by replaying it at `/auth/reset-password` — refused there, and gone from its own flow. That is a denial-of-service on the victim's login, created by the fix. Stores implementing `MagicLinkTokenPeeker` check first; stores that cannot degrade to the after-the-fact check, with the cost stated ("at the cost of burning a misdirected token") rather than hidden.

**Untagged payloads are refused, not grandfathered.** A token minted before this change carries no evidence of its flow, and accepting it for compatibility would hold the confusion open for exactly its TTL — up to 24h. Refusing is the right call and the comment says why.

`errTokenPurpose` is also answered exactly like an unknown token: telling a holder which flow their token *does* belong to is free reconnaissance.

And `MagicLinkTokenPurger` closes a second re-takeover path: completing a reset revoked the user's sessions but left every OTHER outstanding reset link spendable, so a link phished an hour earlier set the password again after recovery.

## What this means for sequencing

I have verified four categories on this PR by execution (body caps, errleak, fieldtypeswitch defaults, analyzer teeth) and none of that touches this code. `battery/auth` is 35 files and +2204 lines, and I have read one of them.

An account-takeover fix should not wait behind a review I cannot complete alone — and it should not merge unreviewed either. That is a real tension and it is worth deciding deliberately rather than by default.

DonaldMurillo:
**Finding on the auth fix itself — the flow separation is half-tested, and the untested half is the one hosts rely on.**

I reversion-tested `redeemPurposeToken`. It has two checks, and only one is guarded:

| mutation | result |
|---|---|
| remove the **peek-time** purpose check | `TestMagicLinkTokenCannotResetPassword` **fails** — caught |
| remove the **post-redeem** purpose check (`strings.CutPrefix` + `errTokenPurpose`) | **entire `battery/auth` suite stays green** |

Deleting the post-redeem check is deleting the flow separation for every store that does not implement `MagicLinkTokenPeeker` — and nothing notices.

## Why that is the half that matters

The code's own comment is explicit that the fallback is the load-bearing guarantee:

> *A store without `MagicLinkTokenPeeker` cannot offer that, and the flow separation is the more important of the two guarantees, so the check still happens — after the fact, at the cost of burning a misdirected token.*

And the design note says tagging lives in this package rather than the store interface precisely so that **"a host wiring its own `MagicLinkTokenStore` gets the separation without implementing anything new."**

So the intended beneficiary of the post-redeem check is exactly the host with a custom store — and that path has no coverage. Both shipped stores (`MemoryMagicLinkTokenStore`, `SQLMagicLinkTokenStore`) implement `PeekToken`, so every test takes the peek branch and the fallback is never exercised.

If the post-redeem check regresses, the verification-token-resets-password takeover reopens **for custom-store hosts only**, silently, with the suite green.

## Fix

A test store that deliberately does **not** implement `MagicLinkTokenPeeker`, running the same cross-flow redemption cases. It should also pin the documented cost — that a misdirected token is burned on that path — so the degradation is asserted rather than described.

That would make mutation B red, and it is a handful of lines given `token_purpose_security_test.go` already has the scenarios.

## Note on how this was found

Not by reading — the code reads correctly, the comments are accurate, and the suite is green. It took reverting each check separately. Same shape as the fixtures that hand-wrote the attribute they were meant to verify, and the post-navigation sequence that ran without asserting: **the scenario is exercised, just never observed.**

Issue #318, opened 2026-08-30
pack drops middleware, plugins and helpers: the round-trip invariant its own banner asserts is unenforced


DonaldMurillo:
## What

`cmd/gofastr/pack.go` opens with a banner stating that `encodeBlueprintYAML`
is "the exact inverse of `decodeBlueprint` (modulo comments + formatting)",
and that when a new blueprint construct is added, **both** the decoder and the
serializer must learn it or the round-trip test fails.

That is false today. `blueprintToMap` never serializes `middleware`, `plugins`,
or `helpers`, even though `topLevelOrder` lists all three. A blueprint carrying
any of them decodes, encodes, and comes back missing them.

## How it survived

`TestPackRoundTrip` uses Meridian, and Meridian happens to declare none of the
three. So the one fixture guarding the invariant is the one fixture that cannot
observe it failing.

Decoding every `examples/*/gofastr.yml` and re-encoding shows **6 of 7 fail** a
strict decode→encode→decode comparison, e.g.:

```
blueprint.Middleware: len 1 != 0
```

## Why it matters

`gofastr pack` is the snapshot tool: it is supposed to produce a `gofastr.yml`
that faithfully describes the app it packed. An app using middleware, plugins,
or helpers gets a snapshot that silently drops them. Regenerating from that
snapshot yields an app missing configuration the original had, with nothing
reporting the loss.

## Suggested shape

1. Teach `blueprintToMap` the three constructs, mirroring the decoders.
2. Make the round-trip test iterate every `examples/*/gofastr.yml` rather than
   one fixture, so a construct the serializer has not learned fails loudly.
   That second half is the durable part — the banner already asserts this
   invariant and nothing enforced it.

Worth checking whether anything else in `topLevelOrder` is likewise listed and
never written; the two lists should be derivable from one another rather than
maintained in parallel.

## Provenance

Found while auditing `pack.go` for #136 (see #317, which fixes a separate
key-injection bug in the same file). Deliberately not folded into that PR: it
is a different defect with a different fix, and the key-safety change should
not carry a serializer feature addition.

DonaldMurillo:
## The durable half of this fix is the test, not the three constructs

Recording the shape before anyone writes it, because the obvious implementation fixes the symptom and leaves the cause.

[...]

PR #343, opened 2026-08-31
pack: make the construct guard reach nested structs

DonaldMurillo:
First item of the 0.77.1 batch, from the pre-release audit's Tier-B list.

## The guard I said was closed wasn't

`TestPackSerializerCoversEveryConstructField` exists to catch a construct that is decoded into the struct and then silently dropped by the serializer — the #318 bug, where `middleware`/`plugins`/`helpers` vanished on pack.

I extended it in #330 to recurse into **slice elements**, watched it catch two live instances (seed `count`/`weights`, entity `renames`), and called the class closed. It never reached through **pointers** or into **nested structs**.

Deleting any one of these single emission lines from `pack.go` left the entire pack + kiln + generator surface green:

| field | before | after |
|---|---|---|
| `scope.cross_owner_read` | green | **red** |
| `scope.tenant_field` | green | **red** |
| `scope.multi_tenant` | green | **red** |
| `pagination.max_list_limit` | green | **red** |
| `auth.jwt_secret` | green | **red** |
| `pwa.short_name` | green | **red** |
| `pwa.background_color` | green | **red** |

**`auth.jwt_secret` is the one with teeth.** A serializer silently dropping an app's JWT secret on `gofastr pack` is precisely the data loss this guard was built to prevent, and the guard could not see it.

Four other fields — `admin.seed_password`, `access.delete`, `relations.unrestricted`, `properties.values` — *did* go red, but only through `TestPack_SerializerRoundTrip`: that is, only because a committed example happens to use them. Coverage by accident, not by construction. That is the same fragility that allowed #318 in the first place, one level further down.

## Why nested structs need a different probe

They can't be probed the way slice elements are. A nested container only emits when a gate sibling is set — `pwa` needs `Enabled`, screen access needs `Auth` or `Role`, `auth` needs any of four fields — so probing `jwt_secret` alone never opens the container, and a guard built that way would fail a *healthy* serializer.

Nested specs invert it: set every leaf so the container is guaranteed open, then zero exactly one field per probe. The zeroed field's absence has to be visible in the output; a dropped emission line makes the probe identical to base, and that identity is the red.

Getting this backwards is the trap — comparing a nested field against a wholly zero blueprint passes for the wrong reason.

## Verification

Each of the seven lines deleted in turn, guard run, red observed, `pack.go` restored and confirmed byte-identical. **7 of 7 caught, where 0 of 7 were before.** Baseline green, full `cmd/gofastr` suite green (257s), `go vet` and `gofmt` clean.

**No exemption was added to make anything pass.** The two that exist are pre-existing (`entities.Endpoints`, `indices.Expression`), each carries its reason in situ, and each still asserts its omission is *total* — a partial leak still fails rather than hiding under the exemption.

## No changelog entry

Test-only; no shipped behaviour changes. The seven fields were already emitted correctly — what was missing was anything that would notice if they stopped.

## Summary by CodeRabbit

* **Tests**
  * Expanded automated coverage for packaging and serialization scenarios.
  * Added validation for nested structures, lists, pointers, and shared ordering data.
  * Improved checks across authentication, administration, progressive web app, access, scope, pagination, exposure, read-scope, and filtering configurations.
  * Added safeguards to ensure configuration data is fully preserved and correctly represented in generated packages.
  * Documented remaining coverage limitations for future improvements.

### DIFF
--- a/cmd/gofastr/pack_test.go
+++ b/cmd/gofastr/pack_test.go
@@ -441,6 +441,26 @@ func TestPackSerializerCoversEveryConstructField(t *testing.T) {
 	}
 }
 
+// KNOWN FRONTIER — what this guard still cannot see, measured, not guessed.
+//
+// The extension above reaches slice elements and one level of nested struct.
+// It does NOT reach structs nested inside a block's props. Deleting any of
+// these single emission lines from pack.go leaves the whole pack + kiln +
+// generator surface green today:
+//
+//	pwa.enabled
+//	p.eyebrow, p.title, p.subtitle          (block props)
+//	src.entity, src.agg, src.field          (props source)
+//	plan.name, plan.price, plan.period      (props pricing plan)
+//
+// None is a live bug: every one of those keys is emitted correctly right now.
+// What is missing is anything that would notice if it stopped — the #318
+// shape, two levels further down than #330 reached.
+//
+// This list exists because the previous extension was declared to close the
+// class and did not. If you extend the guard again, re-derive this list by
+// deleting emission lines rather than trusting it, and shrink it here.
+
 // constructOmissions names construct fields that deliberately do NOT reach
 // the serializer output, with the reason. An omission listed here is part
 // of the contract; an omission not listed here fails the guard above.
--- a/cmd/gofastr/pack_test.go
+++ b/cmd/gofastr/pack_test.go
@@ -338,43 +338,105 @@ func TestPackSerializerCoversEveryAppField(t *testing.T) {
 }
 
 // TestPackSerializerCoversEveryConstructField extends the two guards above
-// one level down: into the structs inside each slice-of-struct construct.
-// The top-level guards prove a construct KEY appears when the Blueprint
-// field is set; they cannot see fields dropped inside the element, because
-// the key's mere presence already differs from the zero output. That is the
-// seed.Count/Weights shape exactly (#330): decoded into BlueprintSeedEntity,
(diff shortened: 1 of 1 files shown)


DonaldMurillo:
Reviewed this myself rather than merging it on the strength of its own numbers, and found the thing I should have looked for last time: **the frontier**.

The seven fields in the description are fixed and proven. But hunting past them — deleting every other single-key emission line in `pack.go` and running the pack surface — turned up **ten more the extended guard still cannot see**:

```
pwa.enabled
p.eyebrow, p.title, p.subtitle       (block props)
src.entity, src.agg, src.field       (props source)
plan.name, plan.price, plan.period   (props pricing plan)
```

Deleting any one of those leaves the whole pack + kiln + generator surface green. Structs nested inside a block's **props** are two levels below where this pass reaches.

None is a live bug — every one of those keys is emitted correctly today. What is missing is anything that would notice if it stopped, which is exactly the #318 shape, two levels further down.

I am **not** expanding this PR to chase them. It moves 0/7 → 7/7 on the fields that include `jwt_secret`, and that is a complete unit. Extending to props-nested structs is a different recursion and deserves its own change.

What I did instead is put the list **in the test** (`4fdc4d84`), with instructions to re-derive it by deleting emission lines rather than trusting it, and to shrink it there when someone extends the guard again.

That is the part that was actually missing in #330. I extended that guard, watched it catch two real bugs, and declared the class closed — while it was blind to the seven fields this PR fixes. The recursion was never the hard part. Not knowing where it stopped was.

Issue #342, opened 2026-08-31
ui-quality-evals is a flaky blocking check: TestMobileCaptureUsesTouchPortraitDPRAndFullDocument times out installing the network guard


DonaldMurillo:
## What

`browser e2e · ui-quality-evals` is a **blocking** required check and it fails
intermittently. It reddened `main` three times in one night, on three unrelated
commits, and it is in the release gate's required set — so it can block a tag at
random.

```
--- FAIL: TestMobileCaptureUsesTouchPortraitDPRAndFullDocument (45.04s)
    evalrunner_test.go:776: mobile capture issues:
        [home mobile: install network guard: context deadline exceeded]
```

Failing runs, all the same job and the same test, nothing else red in any of
them:

| commit | run |
|---|---|
| `7459b226` | 33364983968 |
| `de71ade2` | (same job, earlier run) |
| `1ce802bb` | 33403628906 — blocked the v0.77.0 tag |

## It is also non-deterministic locally

This is the part that matters for whoever picks it up. On the same machine, on
the same commit, with **identical code**:

- one run of `nowarm + 6s budget` **passed** in 3.9s
- a later run of the same configuration **failed** in 8.0s
- an unmodified `main` baseline at the stock 45s budget passed once and failed
  once, twenty minutes apart

So it is not purely a CI-slowness artifact, and a fix cannot be validated by
running the test a few times — that is how you convince yourself of a wrong
theory. It needs either a deterministic reproduction or instrumentation that
records where the 45s actually goes.

## A theory that looked good and is NOT supported

`capture.go:379` starts a 45s budget on a fresh tab, and the first chromedp call
inside it is the network guard's `fetch.Enable` — which is also what lazily
boots the browser target. So the browser launch appears to be billed to a budget
meant for the capture, and the deadline surfaces pointing at the guard rather
than at startup.

Plausible, and I could not support it:

- moving the boot out via `chromedp.Run(browser)` before the budget made it
  **worse** — 46s and failing, where removing that line passed in 3.9s
  (`chromedp.Run` with no actions does not do what the theory needs)
- warming the tab instead, with `chromedp.Run(tab, Navigate("about:blank"))`,
  also failed
- and the run that "proved" the baseline healthy in one experiment failed in the
  next

Recording it so the next person does not spend the same hour on it. The
mechanism may still be right; the evidence is not there.

## Worth deciding separately

Should a **UI quality eval** be a blocking release gate at all? It drives real
[...]

PR #346, opened 2026-08-31
ci: make ui-quality advisory until it is deterministic

DonaldMurillo:
Refs #342.

## Why

`browser e2e · ui-quality-evals` failed on main **three times in one night, on three unrelated commits** — always the same test, always `install network guard: context deadline exceeded` at ~45s, while passing in 15s locally. As a *required* check, each failure held a release tag hostage; v0.77.0 lost an hour to it.

It is also **non-deterministic locally**: identical code passed and failed on consecutive runs. That is the part that decided this. I tried two fixes and the experiment refuted both — the first made the test slower and failing where removing it passed. With no reproduction, no fix can be validated, and raising the timeout without one just moves the cliff.

## What changed

The shard moves out of the `browser-e2e` matrix into its own job. Not `continue-on-error` — this workflow's own comment calls that the last resort ("rerun the job before reaching for continue-on-error").

The name drops the `(blocking)` suffix and the manifest line goes with it **in the same commit**, mirroring what `upgrade-fixtures` did in the opposite direction when it was promoted.

## It keeps running

What it guards is real. The capture test asserts an eval's evidence is genuinely mobile — 1170×2532 at DPR 3, touch points present, portrait orientation, a `Mobile` user-agent, and a full-page shot taller than the viewport. A silently degraded capture would score every future eval against fake evidence.

That is worth seeing. It is not worth blocking a tag while it is this unreliable.

It is also a second opinion rather than the only one: the UI quality the eval scores is covered in part by `browser e2e · site-axe`, the uihost strict checks, and the zero-app-CSS contract (hard rules 7–9).

## Proof the sync guard still holds

Both directions watched:

- advisory job + manifest line removed → `TestReleaseManifestMatchesCI` **passes**
- shard put back in the blocking matrix while the manifest still omits it → **fails**

So this can't drift silently, which is the property that made the `chromium-ui` addition safe earlier today.

## Promotion path

Documented in the job comment: add `(blocking)` back to `name` and the check back to `scripts/release-required-checks.txt` in one commit, once #342 makes it deterministic.

## Summary by CodeRabbit

* **CI and Quality**
  * Moved UI quality evaluations into a separate advisory check.
  * UI quality evaluations no longer block browser end-to-end checks or release publication.
  * Retained the existing evaluation coverage and setup steps.


DonaldMurillo:
Grouped #348's durable half into this PR — same failure class, and this PR is the thing that triggers it.

`scripts/check-branch-protection.sh` (+ `make check-protection`) diffs GitHub branch protection against `release-required-checks.txt`. Run against main right now it reproduces #348's table from live state rather than by hand:

```
    add:    browser e2e · chromium-ui (blocking)
    add:    historical upgrade fixtures (blocking)
    remove: browser e2e · ui-quality-evals (blocking)
```

The `remove` line is this PR: the moment the rename lands, protection requires a context that will never report again and every PR hangs — the PR #193 failure exactly. **So the settings change has to happen before this merges, not after.**

It fails closed. 403 (no admin scope) and 404 (no protection) both mean *cannot verify*, and both are errors — reporting agreement it never checked is the whole failure mode. That is also why it is a make target and not a CI step: the default `GITHUB_TOKEN` cannot read protection. Whether to add a job with an admin token is the open question #348 raised; this leaves that decision open while making the drift detectable today.

Guards mutation-proved — making the unreadable case pass, checking either direction alone, and dropping the empty-manifest refusal each fail exactly the test that names them:

| mutation | fails |
|---|---|
| unreadable protection → pass | `TestBranchProtectionUnreadableFails` |
| only manifest→protection | `TestBranchProtectionOrphanedContextFails` |
| only protection→manifest | `TestBranchProtectionMissingContextFails`, `…EmptyContextsFails` |
| empty manifest accepted | `TestBranchProtectionEmptyManifestFails` |

Also rebased onto main to clear the conflict with the `-p 1` chromium-ui fix (#351); both changes are kept.

DonaldMurillo:
@coderabbitai review

This PR was rate-limited when it was first opened and has had two commits since, so nothing here has been reviewed yet. Please pay particular attention to `scripts/check-branch-protection.sh` and `cmd/gofastr/branch_protection_test.go` — the script fails closed on an unreadable protection API, and the test drives it against a stubbed `gh`.

DonaldMurillo:
Superseded by the #342 root-cause fix shipped in v0.78.0 (#366): the ui-quality flake was the browser launch being billed to the first screenshot's 45s capture budget — the first shot now gets 150s and later shots keep 45s, so the check stays blocking and is believed deterministic. This PR was the other fork of #342's 'advisory or deterministic' choice, and it could not merge anyway without branch-protection edits that are shelved (#348). The check-branch-protection.sh drift checker lives on this branch if that concern ever comes back. Reopen if ui-quality flakes again post-v0.78.0 — that would refute the launch diagnosis and make the advisory move the right call after all.

Issue #339, opened 2026-08-31
widget: a failed chrome fetch after dropping inlined chrome leaves a silent no-op click


DonaldMurillo:
## What

When a ctx-carrying trigger drops SSR-inlined chrome and the replacement fetch then fails, the user gets **nothing** — and no signal that anything went wrong.

Reproduced with a chromedp probe: deep-link arrival → hydrate → close → chrome endpoint returns 500 → click a ctx trigger. Result: `nodes=0 ctxmarks=0 mounted=false`. The inlined node is gone, the fetch failed, and `catch (_) {}` swallows it. The click does nothing at all.

## Why it exists

Introduced by the #321 follow-up in PR #338, deliberately. Before that fix the same interaction showed the **wrong entity's** dialog, which is worse — so this is a real improvement, not a regression. But "nothing happens" is a silent failure, and silent is the property worth removing.

## Scope

Needs all three to line up: a widget inlined via `Hidden().DeepLink(...)`, a ctx-carrying trigger for the same widget on that page, and a chrome fetch that fails. Not common, and not reachable at all for widgets that are never inlined.

## Fix shapes

1. **Re-insert the removed node on failure.** Keep the reference, restore it if the fetch rejects. Returns the user to the pre-fix state — stale chrome rather than none — which is a downgrade in correctness for an upgrade in visibility, so it is a real trade rather than an obvious win.
2. **Surface the failure.** The runtime already has a toast path for a dead RPC (`#fui-nav-toast`); a failed chrome fetch could use it. Does not resurrect the dialog, but the click stops being a no-op.
3. **Retry once before giving up**, then fall back to (2).

(2) is the smallest honest one: it does not pretend the fetch succeeded and does not re-show chrome for an entity the trigger did not name.

## Not urgent

The failure needs a broken chrome endpoint, which is already visible through other symptoms. Filing so the silence is a decision rather than an oversight.

## Provenance

Found by an independent review of #338 via a browser probe, not a failing test. Nothing covers the fetch-failure path after a removal today — that gap is the other half of this ticket.

PR #350, opened 2026-08-31
widget: one toast per failed chrome fetch, not one per click

### LINKED ISSUE #347: widget: a double-clicked trigger raises two toasts for one failed chrome fetch
## What

Clicking a `data-fui-ctx` trigger twice before its chrome fetch resolves produces **two** error toasts for **one** request.

Measured with a browser probe against the slowfail gate: two clicks on the same ctx trigger before resolution → the server saw **1** fetch for `slowfail`, the browser showed **2** toasts.

## Why

Both clicks `await` the same cached promise (`core-ui/runtime/src/widgets.js:102-107`), so the fetch correctly dedupes. But each awaiting call has its own `catch`, and each one runs — so the deduplication that covers the *request* does not cover the *reporting*.

## Severity

Cosmetic. The message is accurate, there is just one too many of it, and it needs an impatient double-click plus a failing endpoint to reach. Filed rather than fixed inside #345 because that PR was already carrying a Medium fix and this is not in its path.

## Possibly pre-existing elsewhere

The RPC toast path (`src/rpc.js:112`) may have the same shape — a shared in-flight promise with per-caller error reporting. **Unverified.** Worth checking before fixing this one in isolation, so the fix lands as a pattern rather than a patch: fixing the widget path while an identical RPC path stays is the sibling-survives failure this repo keeps hitting.

## Fix shape

A per-key "already reported this failure" guard, so the first catch toasts and later ones stay quiet. Keyed on the same `(name, ctx)` the cache uses, cleared when the entry is deleted on failure.

## Provenance

Found by an independent review of #345, by execution rather than reading. Nothing covers repeat-click behaviour on a failing fetch today — that gap is the other half of this ticket.

DonaldMurillo:
Closes #347.

## The bug

Two clicks on a `data-fui-ctx` trigger before its chrome fetch resolves produced **two** error toasts for **one** request. Both clicks await the same cached promise, so the *request* dedups correctly — but each awaiter then ran its own `catch`, and the *reporting* did not dedup with it.

## The fix

Report from the fetch's own `catch`, which runs exactly once per request.

That catch already holds `cache` — the cache object the fetch started against — which is the same navigation signal the awaiter path was keying on via `openedAgainst`. So the #345 post-navigation suppression comes along unchanged rather than being reimplemented in a second place.

The awaiters' catch stays and is silent. It deliberately does **not** rethrow: that would raise an unhandled rejection for a failure already reported once.

## Proof

The dedup needed its own case, because the existing guards cannot see it — with reporting moved back to the awaiters, `TestWidgetChromeCtx_FailedFetchSurfacesToast` (#339), the happy-path-quiet assertion, and the post-navigation suppression (#345) **all still pass**. Only the new assertion reds:

```
two clicks on one failing fetch produced 2 toast(s), want exactly 1:
the request dedups on the cached promise, so the reporting must too
```

With the fix: 1 toast. Full `ChromeCtx|Composed|Budget` family green.

## Sibling check, from the ticket

The RPC path was swept before touching this and does **not** share the shape:

- `_dispatchRPC` uses per-signal abort controllers, so a rapid double-click aborts the prior request — and it already filters `AbortError` (`rpc.js:284`). Without that it would report a failure for a request the user themselves superseded.
- `_submitForm`'s catch has no abort filter, but its fetch carries no `signal`, so no `AbortError` can reach it. Each submit is its own request: a double-submit is two real failures and two accurate messages.

So this fix is correctly local, not one half of a pattern.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate error notifications when multiple actions share the same failed request.
  - Ensured failed widget loads display a single toast after the request completes.

- **Tests**
  - Added coverage confirming that repeated clicks during one in-progress request produce only one failure notification.

### DIFF
--- a/core-ui/runtime/src/widgets.js
+++ b/core-ui/runtime/src/widgets.js
@@ -99,10 +99,6 @@
     // cache. '\0' cannot survive HTML attribute parsing, so name+ctx
     // pairs cannot collide.
     const key = name + '\0' + (ctx || '');
-    // Cache identity for THIS open. gofastr:navigate swaps NS._chromeCache
-    // for a fresh object, so an inequality later means the user left the
-    // page this click belonged to.
-    const openedAgainst = NS._chromeCache;
     const cached = NS._chromeCache[key];
     if (cached) {
       // LRU refresh: delete + re-insert moves the key to the recency
@@ -127,6 +123,22 @@
           return await r.text();
         } catch (err) {
           delete cache[key];
+          // #347: report HERE, not in the awaiters' catch below. Every
+          // click that joined this in-flight promise runs its own catch,
+          // so reporting there gave one toast per click for a single
+          // failed request — the dedup covered the fetch and not the
+          // telling. This runs exactly once per fetch.
+          //
+          // #339/#345: only if the user is still on the page they
+          // clicked on. `cache` is the object this fetch started
+          // against; gofastr:navigate swaps NS._chromeCache, so a
+          // mismatch means they walked away — and in the worst case
+          // they re-opened the panel successfully after navigating, and
+          // the toast would deny an open panel.
+          console.error('[gofastr] widget chrome fetch failed', err);
+          if (cache === NS._chromeCache) {
+            NS._toastOrFallback?.({ variant: 'error', title: 'Could not open that panel.', ttl: 6000 });
+          }
           throw err;
         }
       })();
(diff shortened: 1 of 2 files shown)


DonaldMurillo:
Reviewed this myself — `omp` dropped off PATH mid-session so I could not dispatch an independent reviewer, and I would rather say that than imply this got one.

I went after the risk that actually matters when you move a report: **did any failure path become silent?** `#339` existed to delete a silent no-op, so re-introducing one here would be the worst possible regression.

Three paths, worked through:

**LRU eviction stranding an awaiter — not reachable.** I expected this to be the hole. `NS._chromeCache[key]` is read *synchronously* on the line following the insertion block, with no `await` between them, so no other click can interleave and evict the key first. A later call aging this key to `ks[0]` cannot affect a read that already happened. If an `await` is ever introduced between those two points, this stops being true and `await undefined` becomes a silent no-mount.

**Cache-hit joiners staying quiet — correct, not a lost case.** A joiner takes the `if (cached)` branch and awaits a promise whose *creating* call owns the inner catch. That catch reports once. The joiner's silence is the fix, not a gap.

**Unhandled rejection — none.** The creating call attaches its `await` synchronously in the same run, so the rejected promise always has a handler. The rethrow is kept because it preserves "this promise failed" for any future awaiter; without it the IIFE would resolve to `undefined` and the awaiter's catch would be dead code.

What I did **not** do: re-run the both-ways vacuity check an independent reviewer would have. I ran it once when writing the fix — reporting from the awaiters gives `2 toast(s), want exactly 1` and the three pre-existing guards stay green either way — but I have not had someone else reproduce that.

PR #352, opened 2026-09-01
test(auth): cover the token-purpose check on stores that cannot peek

DonaldMurillo:
Follow-up to #341. The token-purpose fix landed with a check that no test could fail.

`redeemPurposeToken` checks a token's purpose twice — once via `PeekToken` before consuming, and again on the payload `RedeemToken` returned. `MemoryMagicLinkTokenStore` and `SQLMagicLinkTokenStore` both peek, so every test in `token_purpose_security_test.go` takes the first branch. Deleting the second check leaves the entire `battery/auth` suite green.

That is the check the design rests on. From the doc comment on the file: tagging lives in `battery/auth` rather than in `MagicLinkTokenStore` so "a host wiring its own MagicLinkTokenStore gets the separation without implementing anything new" — and a host implementing the bare three-method interface reaches only the post-redeem check. Had it regressed, verification-token-resets-password reopened for custom-store hosts, silently.

## What this adds

`nonPeekingTokenStore` forwards to a SQL store by hand rather than embedding one, so `PeekToken` and `DeleteTokensForPayload` are not promoted; the fixture calls `t.Fatal` if either interface is ever satisfied, because that is how this test quietly stops testing anything. The four cross-flow scenarios run against it, plus one more:

`TestNonPeekingStoreBurnsMisdirectedToken` pins the documented cost of the fallback. `RedeemToken` is atomic, so on that path the token is already consumed when the purpose is checked — anyone holding a victim's magic link can burn it by replaying it at `/auth/reset-password`. `TestMagicLinkTokenCannotResetPassword` asserts the opposite (302) on a peeking store. The contrast is the argument for implementing `MagicLinkTokenPeeker`, and it is now in the suite instead of only in a comment.

`newTokenPurposeHarness` gained a variant taking the store; the existing tests call through it unchanged.

## Proof

Mutation, both directions:

| mutation | result |
|---|---|
| remove the post-redeem check | 4 new tests fail, nothing else |
| remove the peek-time check | only `TestMagicLinkTokenCannotResetPassword` fails |
| as written | `battery/auth` green |

Neither mutation is masked by the other, so each check has its own test.

### DIFF
--- a/battery/auth/token_purpose_nopeek_test.go
+++ b/battery/auth/token_purpose_nopeek_test.go
@@ -1,0 +1,151 @@
+package auth
+
+// Cross-flow token purpose binding, on a store WITHOUT MagicLinkTokenPeeker.
+//
+// redeemPurposeToken checks the purpose twice: once at peek time, and again
+// on the payload RedeemToken returned. Every store shipped in this package
+// peeks, so token_purpose_security_test.go only ever exercises the first
+// check — removing the second leaves that whole file green.
+//
+// The second check is the one the design note is about: tagging lives in
+// this package rather than in MagicLinkTokenStore so "a host wiring its own
+// MagicLinkTokenStore gets the separation without implementing anything
+// new". A host that implements the bare three-method interface is exactly
+// the case the post-redeem check exists for, and it had no test.
+//
+// These run the same cross-flow scenarios against nonPeekingTokenStore, and
+// pin the documented cost of that path: a misdirected token is burned.
+
+import (
+	"context"
+	"net/http"
+	"testing"
+	"time"
+)
+
+// nonPeekingTokenStore is the bare MagicLinkTokenStore a host gets from
+// implementing the documented interface and nothing else. It forwards to a
+// real store by hand rather than embedding it: embedding would promote
+// PeekToken (and DeleteTokensForPayload) and quietly put these tests back on
+// the peeking path.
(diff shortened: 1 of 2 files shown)


DonaldMurillo:
Review notes on this PR, since CodeRabbit is rate-limited on it — its `0 unresolved` here means it reviewed nothing, not that it found nothing.

**Verified while writing this, no bug:** `confirmHandler` (`magiclink.go:397`) peeks directly rather than going through `redeemPurposeToken`, so it is a fourth path that could have skipped the purpose check. It does not — it calls `peekPurposePayload(got, purposeMagicLink)` and 401s on a mismatch, with a comment explaining that the page would otherwise display a reset token's userID as the address to sign in as.

**Design note, not a change here.** The fallback's cost exists only because the purpose lives in the *payload*, which is behind `RedeemToken`. Putting it in the *token string* instead — `createPurposeToken` returning `"pwreset." + token` and `redeemPurposeToken` stripping and checking the prefix before calling the store — would make the purpose checkable with zero store cooperation, on every store, with nothing burned. Payload tagging could stay underneath as defence in depth. I have not checked every call site for assumptions about the token string's shape (the confirmation page passes it through a form field and a URL), so this is a proposal rather than a claim, and it is a change to a shipped security fix — worth deciding deliberately rather than folding into a test PR.

Issue #136, opened 2026-07-25
Security audit coverage: surfaces that are never-looked, not clean


DonaldMurillo:
The v0.43.0 audit (PR #134) covered a lot but did **not** clear everything it touched. `.claude/skills/adversarial-tests/SKILL.md` defines a surface as CLEARED only when the Opus deep pass, the Haiku breadth pass, `go vet` and `govulncheck` are all dry on it. The surfaces below had fewer than four signals, so they are **never-looked, not clean** — this issue exists so that distinction survives.

### Highest value first

- [ ] **`codegen/runner.go` + `codegen/extension_command.go`** — executes external commands from a codegen config file. A genuine exec surface that no pass has opened. Do this one first.
- [ ] **kiln journal replay** — `kiln/journal/replay.go` re-applies raw payloads at boot with minimal checks, skipping `protocol.go`'s semantic guards (plan-gating for destructive ops, the multi-tenant refusal). A hand-authored `.kiln.session.jsonl` can install world state the HTTP API would refuse. No HMAC, no schema-version check, no per-entry integrity. Precondition is local FS write, so severity is low today — but it becomes a real deserialization vector if journals are ever shared or transported.
- [ ] **`cmd/gofastr generate`'s Go emitter** — the tail of the blueprint injection class. PR #134 fixed the e2e-test sinks; the entity/screen emitters' quoting of IR-derived strings was not audited.

### Never opened

- [ ] `core/store`, `core/markdown`, `core/fuzzy`, `core/yaml`, `core/moduleproto` (peer dispatch)
- [ ] `core/middleware/{cors,logging,timeout,metrics,idempotency_sql}.go`
- [ ] `core/mcp/resources.go`, `framework/crud/llmmd.go` (grep only)
- [ ] `framework/processmodule_probe*.go`, `processmodule_tools.go`, `processmodule_broker.go`
- [ ] Most `core-ui/runtime/src/*.js` modules — the breadth pass enumerated their sinks but only `combobox` and `tree` were verified; the rest (carousel, taginput, dropzone, fileupload, menu, popover, …) are unreviewed
- [ ] `core-ui/check`, `core-ui/widget/{theme,preset}`, `framework/uihost/{uinoderender,internal}`
- [ ] `cmd/gofastr/generate_cli.go` (47KB) and `pack.go` (78KB) — greps only
- [ ] `examples/ecommerce` never diffed against a fresh regeneration

### Known non-findings (do not re-file)

[...]

PR #355, opened 2026-09-01
fix(kiln): refuse an empty plan_id, and test the refusal that mattered

DonaldMurillo:
Auditing the **kiln journal replay** item on #136. Most of it is already closed — the semantic guards named in that ticket are present and hold under mutation. One did not.

## The guard nobody had watched fail

`spendPlan`'s `planID == ""` branch is the one carrying the whole plan-gating property, and it had no test.

`plan_proposed` did not require a non-empty id, so a hand-authored `.kiln.session.jsonl` could store *and approve* `s.Plans[""]` listing a target. A destructive entry that simply **omits** `plan_id` decodes to `""`, and the lookup then hits that approved plan. Watched it, rather than reasoned it:

| | result |
|---|---|
| guard present | `delete_entity "orders": no plan_id` — entity survives |
| guard removed | `err = <nil>` — entity gone from the world |

The precondition is a local filesystem write, so this is an integrity property rather than a remote hole — the same framing #136 uses. The point stands either way: the log is supposed to be the authorization record, and one lookup miss away from not being one.

The empty id is now refused at both ends — `plan_proposed` will not store it, `spendPlan` will not spend it.

## Testing the deeper lock needs care

Reaching `spendPlan`'s branch *through* `Apply` is impossible once `plan_proposed` refuses first: the setup never happens, so removing `spendPlan`'s branch fails nothing. That is exactly how the shallower lock hides the deeper one — and how the deeper one went untested in the first place. The subtest plants the approved empty-id plan directly in the session and calls `spendPlan` itself.

## Proof

| mutation | fails |
|---|---|
| `plan_proposed` accepts an empty id | `…/plan_proposed_refuses_an_empty_id`, `…/whitespace_is_not_an_id_either` |
| `spendPlan`'s empty-id refusal removed | `…/spendPlan_refuses_an_empty_id_even_with_an_approved_empty-id_plan` |
| both removed (the pre-fix state) | all five assertions |

Full `./kiln/...` green, including the 162s integration suite.

One note on method, because it nearly cost me this finding: my first proof harness reported "both locks removed" as a **miss**. The harness was indexing the mutation's label instead of its code, so it never mutated anything — a passing test against an unmodified file, read as evidence of a gap. It now asserts the mutation actually applied before running. Second false negative from a proof harness this session; the first was a `-run` pattern that matched zero tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented destructive operations from being authorized with missing or whitespace-only plan IDs.
  * Forged journal entries that omit a plan ID no longer delete data.
  * Invalid plan proposals are rejected and are not stored.

* **Tests**
  * Added coverage for empty and whitespace-only plan IDs, including end-to-end protection against unauthorized deletion.

### DIFF
--- a/kiln/journal/replay.go
+++ b/kiln/journal/replay.go
@@ -77,6 +77,17 @@ func Apply(s *Session, e Entry) error {
 		if err := e.Decode(&p); err != nil {
 			return err
 		}
+		// A plan with no id is not a plan, and storing one arms the only
+		// trick that gets past spendPlan's empty-id refusal: propose and
+		// approve s.Plans[""] listing a target, then emit the destructive
+		// entry with plan_id OMITTED, which decodes to "". Without
+		// spendPlan's first branch that lookup HITS an approved plan
+		// listing the target and the deletion is authorized. That branch
+		// holds today, so this is the second lock on the same door — but
+		// it removes the state that makes the door interesting.
+		if strings.TrimSpace(p.PlanID) == "" {
+			return fmt.Errorf("plan_proposed: empty plan_id")
+		}
 		s.Plans[p.PlanID] = &Plan{
 			PlanID:     p.PlanID,
 			ProposedAt: e.Timestamp,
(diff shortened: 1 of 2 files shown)


DonaldMurillo:
Batched into #361 with the other two audit fixes — one CI run instead of three. Same commit, unchanged.

PR #359, opened 2026-09-01
fix(middleware): a non-positive Timeout means no deadline, not an instant one

DonaldMurillo:
Found auditing #136's never-looked list. Three things, one package.

## `Timeout(0)` 504s every request instead of meaning "no deadline"

`Timeout` never guarded its duration, so `time.NewTimer(0)` fires before the handler can run. Measured:

| call | result |
|---|---|
| `Timeout(0)` | 504 on 50 of 50 instant requests |
| `Timeout(-1s)` | 504 on 50 of 50 |
| `RouteTimeout{Budget: 0}` | 200 |

Same value, same file, opposite meanings. The route path was already right and says so in its own comment — "a negative Budget skips the deadline entirely, matching net/http where a zero duration means no timeout". The constructor now reads it the same way.

The trap is a config field left unset: `Timeout(cfg.RequestTimeout)` with `RequestTimeout` zero turns the surface **off** rather than leaving it untimed, and it fails closed on every request rather than loudly at startup. Nothing in-repo tripped it because `app.go` installs the middleware only when `RequestTimeout > 0` — the cover is the framework's wiring, not the API, and a direct caller of the core package has no such cover.

Mutation-proved: removing the guard fails `TestTimeoutZeroOrNegativeMeansNoDeadline` on both the `0s` and `-1s` cases.

## Deadline branch tie-break — kept, but not proven, and the test says so

A `select` whose cases are both ready picks between them uniformly at random, so a handler that *returns* inside its budget can still 504 when the middleware goroutine is descheduled past the deadline. The branch now re-checks `done` before abandoning.

**This ships without a gate.** Two reasons, both in the test comment:

1. No deterministic version exists. A select picks randomly only when both channels are ready at the moment it is *evaluated*; enter it earlier and it blocks and wakes on the first one. Producing the overlap needs a deschedule nothing outside the runtime can arrange.
2. I could not reproduce the failure. With the re-check removed, the load probe found **0 misfires in 4000 iterations under 10 CPU burners** (~50 genuine overruns per 2000, so the deadline was firing plenty). The audit worker reported ~2 per 2000 on the same shape. Not reproduced here.

The tie-break is kept because the branch is reachable by construction — the Go spec makes the ready-vs-ready choice uniform — and because it cannot make anything worse. Not because a failure was observed. That distinction is written into the test rather than left to a commit message.

The probe also carries the trap that cost my first draft its answer: the 504 path returns while the handler is still running, so reading its elapsed time straight after `ServeHTTP` reads zero for exactly the requests under test. Counting those as overruns makes the instrument blind to what it is looking for — fixing it moved ~250 of every 300 "overruns" per 2000 into their real bucket.

## `IdempotencyStore.Begin` doc had the wrong convention

It said "replay nil, ok=true: the caller is the first writer". Both shipped stores return `ok=false` there (`idempotency_sql.go:168-171`, `idempotency.go:498`) and the middleware's replay branch is `ok && replay != nil`, so `ok=false` is the contract and always was. Implementations and caller agreed with each other; only the comment was wrong. A store author following it would ship `ok=true`, which happens to still work — so the cost is confined to misleading the next implementer.

## Verification

`go test ./core/middleware/ -count=1` green (9.0s); `go test -race ./core/middleware/ -count=1` green (24.7s).

### DIFF
--- a/core/middleware/idempotency.go
+++ b/core/middleware/idempotency.go
@@ -29,8 +29,12 @@ const IdempotencyKeyHeader = "Idempotency-Key"
 //   - replay non-nil, ok=true: a cached response already exists for this
 //     key and fingerprint; the middleware should write replay back and
 //     skip the downstream handler.
-//   - replay nil, ok=true: the caller is the first writer for this key.
-//     It must call Finish exactly once with the captured response.
+//   - replay nil, ok=false, err=nil: the caller is the first writer for
+//     this key. It must call Finish exactly once with the captured
+//     response. (Both shipped stores return ok=false here, and the
+//     middleware's replay branch is `ok && replay != nil`, so ok=false is
+//     the contract. This line used to say ok=true, which no implementation
+//     and no caller has ever agreed with.)
 //   - ok=false, err=ErrFingerprintMismatch: same key was used previously
 //     with a different request fingerprint. The middleware responds 422.
 //   - ok=false, err=ErrInFlight: another request with the same key is
--- a/core/middleware/timeout.go
+++ b/core/middleware/timeout.go
@@ -281,6 +281,20 @@ func Timeout(d time.Duration) Middleware {
 				}
 				effective = rt.Budget
 			}
+			// Same rule for the constructor's own duration as for a route
+			// budget, and for the same reason: zero means no deadline, as
+			// it does in net/http. Without this, time.NewTimer(0) fires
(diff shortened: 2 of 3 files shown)


DonaldMurillo:
Batched into #361 with the other two audit fixes — one CI run instead of three. Same commit, unchanged.

PR #396, opened 2026-09-06
feat(i18n): per-page document language, translatable pager, structured fence info strings

DonaldMurillo:
Three gaps a bilingual documentation site runs into. Each was found by building
one on GoFastr, and each is additive: an app that configures none of them
renders byte-identically.

## 1. A per-page document language

`<html lang>` came from one site-wide value, so every page of a bilingual site
claimed the site language. Two things break as a result:

- A screen reader reads the Spanish pages with English pronunciation rules,
  which is **WCAG 3.1.1** failing on every translated page.
- A full-text indexer that picks a page's language by reading `<html lang>`
  (Pagefind does) builds one index for the whole site and stems every language
  with English rules, so per-language chunking never happens.

There was no per-page hook of any kind: `screenHeadHTML` injects into `<head>`
and cannot set an attribute on `<html>`.

Two knobs, because a page's language is known in two different places:

```go
// The common case: a whole prefix in one language.
application.WithLangFunc(func(path string) string {
    if strings.HasPrefix(path, "/es/") {
        return "es"
    }
    return ""
})

// A page that knows its own language, including one that learns it in Load.
func (s *ArticleScreen) ScreenLang() string { return s.article.Lang }
```

`ScreenLanger` is read after `Load`, exactly like `ScreenTitle`, and wins for
that page. `uihost.WithLangFunc` does the same for the shells the host builds
itself (404, 405, embed frame), falling back to the app's `LangFunc` so a site
declares its languages once. Every fallback ends at `EffectiveLang`.

Covers every emit site, including `renderscreen.go`, which the original survey
missed.

## 2. Translatable prev/next pager labels

`DocPrevNext` wrote `← Previous` and `Next →` as literals, with no fields on
`DocPager` to change them, so a non-English docs site was stuck with English
arrows. `PrevDirLabel` / `NextDirLabel` default to those exact strings.

The arrow is part of the value rather than something the component adds, because
a translation may want it on the other side of the word.

## 3. Fence info strings, and fences longer than three

Two CommonMark rules the three-backtick scan got wrong.

**The info string is a language followed by options, not one opaque name.**
Folding the whole line into the class produced
`class="language-go title=&quot;main.go&quot;"`, which matches no language — so
writing any fence option silently cost the block its syntax highlighting.
`ParseFenceInfo` splits it; the options travel on a new `data-meta` attribute,
and `framework/ui.Markdown` maps `title=` / `filename=` to
`CodeBlockConfig.Filename` and `showLineNumbers` to `LineNumbers`, ignoring
anything else.

**A fence is three or more of the same character**, and closes only on a run at
least as long with nothing but whitespace after it. That is the only way to
write a Markdown example containing a ``` block; reading exactly three tore such
an example apart at its first inner fence.

Plain fences stay byte-identical, pinned by `TestPlainFenceOutputUnchanged`
across five shapes.

## Verification

Built a real bilingual site against this branch: the live server serves
`lang="en"` on `/docs/...` and `lang="es"` on `/es/...`, which is what makes
Pagefind report `Discovered 2 languages: en, es` and write a separate chunked
index per language.

Escaping checked on both new paths — a `lang` value containing a quote is
escaped to `&quot;` rather than breaking out of the attribute, and a
`title=""` reaches the filename header escaped.

Fence edge cases verified: tilde fences holding backtick fences, a fence closed
by a longer run, unclosed fences, five backticks holding four, and a tilde fence
not closed by backticks.

`go build ./...` and `go vet ./...` clean. `core-ui/app`, `core/markdown` and
`framework/ui` pass. New tests: 8 in `core-ui/app/lang_test.go`, 3 in
`framework/uihost/lang_test.go`, 3 in `framework/ui/doc_layout_test.go`, 8 in
`core/markdown/fence_test.go`, 3 in `framework/ui/markdown_test.go`, each task
including an explicit unchanged-default test.

Two caveats on a full local run, both pre-existing on this machine and verified
identical at the base commit: `gofmt -l .` lists every file in the repo because
the checkout is CRLF, and `go test ./...` has failures from Windows ACL checks,
`os.Root` TempDir cleanup, sqlite, and Chrome-driven e2e. `framework/docs`
`TestDocExamplesCompile` fails the same way on `origin/main`.

## Summary by CodeRabbit

- **New Features**
  - Added per-page document language resolution, including route- and screen-specific language settings with sensible fallbacks.
  - Added customizable Previous and Next labels for documentation navigation, supporting translations.
  - Added support for code-fence metadata such as filenames and line numbers.

- **Bug Fixes**
  - Improved syntax highlighting when code fences include options.
  - Fixed parsing for longer fenced code blocks and improved handling of backtick and tilde fences.
  - Ensured language attributes are correctly applied to rendered pages and embedded shells.


DonaldMurillo:
@coderabbitai please review

PR #420, opened 2026-09-15
fix(headless): extras and overrides store attribute names folded

DonaldMurillo:
## What

`Safe` and the override sanitiser in `framework/headless` store attribute names folded, the way the browser reads them, and refuse one key given under two spellings.

## Why

CodeRabbit's late pass on #418 found it, outside that PR's diff: `Safe` kept a caller's spelling, so `Extra{"NAME": "caller"}` on an `Input` landed beside the component's own `name`. Sorted output put `NAME` first, and the browser keeps the first attribute it reads, so the submitted field carried the caller's value instead of `InputProps.Name`. Every seam that goes through `Safe` (`ExtraAttrs`, `Extra`, the `Action` seam) and every Override had the same shape.

## How

- `Safe` lowercases each accepted key before storing it, so a later `attrs["name"] = p.Name` or a `Merge` with the component's own keys overwrites the caller's spelling rather than sitting beside it.
- `allowedOverride` does the same for the Overrides seam.
- A key given twice after folding panics at render, the rule `Owned` already applied to `min`/`MIN`.
- `TestExtrasAndOverridesStoreKeysFolded` renders an `Input` with `NAME` in its extras and asserts one `name` attribute, the component's; refuses two spellings through `ExtraAttrs`; refuses two spellings through an Override. Reverting the fold fails the test.
- Docs: the seam paragraph and a "Common mistakes" entry in `ui-headless`, the same entry in `framework/headless/agents.md`, a changelog line.

No golden changed: every key the fixtures pass was already lowercase.

## Summary by CodeRabbit

- **Bug Fixes**
  - Attribute names supplied through headless component extras and overrides are now handled case-insensitively, matching browser behavior.
  - Component-provided attributes take precedence over caller-provided variants with different capitalization.
  - Conflicting attributes supplied under multiple capitalizations are now refused instead of producing unpredictable results.

- **Documentation**
  - Updated headless component guidance and common-mistake examples to explain attribute-name handling and duplicate-key behavior.


DonaldMurillo:
@coderabbitai review

DonaldMurillo:
Folded into #419 as a cherry-pick; one review budget, one PR.

PR #431, opened 2026-09-18
fix(runtime): no module URL is bare, and only a matching ?v= caches forever

AlexCiccolella125:
A registered module the manifest carries no version for was loaded under a bare URL, and the serve route answered every module URL with a year-long immutable `Cache-Control`. Together they pinned such a module in the browser under a URL that can never bust, so the next deploy never reached it. Found while `framework/local`'s modules were being registered through the behaviour seam that DonaldMurillo/gofastr#419 added, where a bare module URL is the common case for a behaviour the host has not versioned. The fix stands on its own and is the first PR of that series.

Two halves. The loader always sends a `?v=`: when the manifest has no version for a module the URL carries a sentinel instead of nothing (`boot.js`, `runtime.js`), which also costs fewer bytes than the ternary it replaces, so the level-1 budget line moves down 4 bytes to keep the cliff fixture's bracket honest and `SYMBOLS.txt` loses the symbol the ternary needed. And the route earns the immutable header only when `?v=` is the served bytes' hash (`core-ui/widget/server.go`); any other request answers `no-cache`, the policy every other `/__gofastr` script already follows.

Review fix, from the maintainer's pass: `ServeComputeAsset`, fifty lines below, had the same bug. The compute manifest is an inline block in the document, so a tab open across a deploy asks for the old hash and was answered with the new bytes marked immutable. It now needs a matching hash too, and a test asks for a matching, a stale and a missing `?v=` and reads the header each time; with the check reverted both mismatch arms fail. A style commit then shortened the comments the fix had added, which the review bot had flagged as restating the code; no token outside a comment changed.

Gates: `core-ui/runtime` (budget, symbols, fragment and composed-runtime gates), `core-ui/widget` and `framework/uihost`'s compute tests pass. The original commit was cherry-picked from `feat/framework-local`, where it has run under the full runtime suite.

## Summary by CodeRabbit

- **Bug Fixes**
  - Runtime modules and compute assets now use versioned URLs consistently.
  - Requests with missing or outdated versions are no longer cached indefinitely, preventing stale code from being reused.
  - Correctly versioned modules and assets continue to receive long-lived immutable caching for faster subsequent loads.

- **Tests**
  - Added coverage for caching behavior with valid, missing, and stale versions across runtime modules and compute assets.


DonaldMurillo (changes requested):
Reviewed on `81d4333` with the repo's own gates plus an independent pass. One change requested, and it is in this PR's own file.

## What I verified

- `go build ./...`, `go vet`, `make analyze` (the full vettool set): clean.
- `go test ./core-ui/runtime/` in full, 384s — the attrdoc, fragment-map, SYMBOLS and budget gates all pass, so the regenerated `SYMBOLS.txt` line and the two edited copies of `loadModule` are consistent. `TestComposedRuntimeMatchesOnDiskFile` is what makes the `frag/boot.js` ↔ `runtime.js` duplication safe to edit by hand; it passes.
- `go test ./core-ui/widget/`: pass.
- **The new guard is not vacuous.** I reverted `serveRuntimeModule` to the unconditional `immutable` and `TestServeRuntimeModuleRefusesToFreezeAnUnversionedURL` fails on the first arm (`no ?v=: Cache-Control = immutable, want no-cache`).
- **The budget bracket is real, and measured rather than argued.** Independent measurement of the two sides: real core at level 1 is 15427, the cliff fixture is 15443, the new window is 15442. That is the largest value below the fixture's crossing, so `TestCoreBudgetRejectsCliffOverflow` and `TestRuntimeModuleSizeBudgets` pin both edges. The comment's claimed bracket is exact.
- The sentinel cannot collide with a real hash: `ModuleHash` returns `hex.EncodeToString(sum[:8])`, always 16 characters, and an unknown name 404s before any header is set.
- `no-cache` on the mismatch arm matches `framework/uihost/assetserve.go:34-40` and the registry-CSS route, both of which have tests pinning "a mismatched `v` must not be immutable". No `Vary` is needed — the query string is already part of the cache key and the body depends only on path plus query.
- The service worker does not re-freeze what the HTTP layer just refused: `pwaPrecachePaths` only ever lists real content hashes and nothing is added to Cache Storage at runtime, so a `?v=x` URL is always a miss and falls through to the network.

## Change requested

**`ServeComputeAsset` is the same bug, 50 lines below the fix, in this file.**

`core-ui/widget/server.go:230` still sends `Cache-Control: public, max-age=31536000, immutable` on every request to `/__gofastr/compute/<name>.js|wasm`, with no check that `?v=` matches anything.

It is narrower than the module case this PR fixes. `compute.js`'s `entryFor` throws before a URL is ever built when the manifest lacks the entry, so there is no bare-URL fallback to exploit. The path that survives is a stale hash: the compute manifest ships as an inline `<script id="gofastr-compute-assets">` block in the document, so a tab left open across a deploy holds the old manifest, requests `?v=<oldhash>`, and gets the **new** bytes back marked immutable. That pins an (old-URL, new-body) pair in the browser for a year, which is the exact hazard `framework/uihost/uihost.go:3980` documents as the reason its own gate exists.

The canonical spelling is `framework/uihost/assetserve.go:34-40`, and `framework/local/manifest.go:120-125` on branch #433 already follows it. The fix here is the three lines this PR just wrote, applied one function down.

I am raising it on this PR rather than as a follow-up because CLAUDE.md asks for the sweep by the invariant rather than by the name, and because the commit message says "only a matching `?v=` caches forever" while one route in the file it edits still does not.

AlexCiccolella125:
Fixed on the branch, three commits on top of the original.

`ServeComputeAsset` now sends the immutable header only when `?v=` is the asset's hash and `no-cache` otherwise, the three lines from one function up. `TestServeComputeAssetRefusesToFreezeAMismatchedHash` asks for a matching, a stale and a missing `?v=`; with the check reverted to the unconditional header, the stale and missing arms fail (`Cache-Control="public, max-age=31536000, immutable", want "no-cache"`). The two existing compute tests now request the real hash instead of the literal `hash`, which had only ever passed because the route ignored it.

A style commit then shortened the comments the fix had added (the loader note in `boot.js` and `runtime.js`, the route comment, the budget line's history), which the review bot had flagged as restating the code. The composed-runtime and budget gates agree that no token outside a comment changed. Thank you for the sweep by invariant rather than by name; the commit message's claim is true of the whole file now.

PR #432, opened 2026-09-18
feat(runtime): the browser's own store, as one primitive, and a slice that remembers

AlexCiccolella125:
## What this is

The runtime had five hard-coded Web-storage keys: a sidebar boolean, a banner bit, the colour scheme, a form draft, a scroll map. Each was written where it was needed, each re-derived its own namespace, and nothing sat under them. An app that wanted to remember anything of its own had to leave the framework, write a document script, and own the DOM from there. This PR is the primitive those five keys should have been sitting on, plus the one thing `core-ui` is willing to say on top of it: a `store.Slice` the browser remembers. It is the floor and nothing more. Schemas, migrations, caps per collection, an upload channel, anything with an opinion about what a stored value means, belong to a layer above `core-ui`, and that layer is DonaldMurillo/gofastr#433, the framework/local PR stacked on this one. Offline-first stays a non-goal, and the capability map says so in its new row.

## The primitive

`core-ui/runtime/src/local.js` is one browser store with one owner. `window.__gofastr.local` is `available`, `get`, `set`, `remove`, `keys`, `entries`, `subscribe` and `watch`.

IndexedDB is the engine, because a saved value is not a preference: it is asynchronous, it is not capped at the few megabytes `localStorage` shares across a whole origin, and a large read does not block the main thread. `localStorage` is the fallback, used only when IndexedDB will not open and only for values up to 8 KiB. Both are browser APIs, so nothing new is depended on.

Every entry in either engine lives under the literal `gofastr.state.` plus the component-encoded key, so an application key can never name another feature's storage. The storage-key lint holds that spelling where the write happens rather than by convention.

`keys(prefix)` and `entries(prefix)` read one IndexedDB key range rather than scanning the store, and an entry carries its stored UTF-8 size. `watch(prefix, fn)` hears another tab's write of any key under a prefix. Those three are what a layer that groups records under a common prefix needs, with no opinion here about what a group means.

Every call settles rather than throwing. `set` names which of `size`, `quota`, `encode` or `unavailable` refused it. Truth still lives on the server. There is no `data-fui-*` attribute and no core-bundle cost; owners reach the module through `registry.Requires("local")` or `__gofastr.loadModule('local')`, the way `action` is reached.

The runtime's five keys moved onto it. `fragments.go` and `runtime-contract.md` carry the API; `core-ui/ARCHITECTURE.md` says where the primitive sits.

## Four things it could not say, or said twice

Found by re-reading the primitive against the layer that consumes it, each its own commit with a test that fails without it.

`available()` could not name the engine that answers. Two booleans told a caller what was possible, not what would serve the next read. It returns `{idb, ls, engine}`, `engine` being `idb`, `ls` or `none`. Test: `TestLocalNamesItsEngineAndSettlesItsEnumerations`.

An aborted enumeration and an empty store were the same answer. `keys()` and `entries()` resolved a bare array, so a `clear()` built on them reported success while every record survived: a logout that says the previous user's data is gone and is wrong. Both settle `{ok, reason, ...}` now, the shape `set` and `remove` already kept. Same test.

The two engines could not see each other. Entries a fallback session wrote were invisible to every IndexedDB session, so a previous user's records outlived a logout. The first session that opens the database adopts them and empties the namespace out of `localStorage`, one way, and `remove` reaches both engines. Test: `TestLocalAdoptsFallbackEntriesWhenIndexedDBReturns`.

The fallback announced a write twice, on BroadcastChannel and as the native `storage` event, so every subscriber in every other tab fired twice for one write. The storage event carries it alone. Test: `TestLocalFallbackDeliversAWriteOnce`.

## `store.Slice.Persist()` and `.PersistMax(n)`

A signal is a UI projection and the server is where truth lives, but some state has no server to live on: a draft, a filter set, a scratch list a signed-out visitor builds in one browser. `Persist()` marks such a slice. Every binding it renders carries `data-fui-signal-persist="<cap>"`, the marker that loads `signal-persist` (`core-ui/store/persist.js`, `Requires("local")`). The module restores the browser's value into the signal after hydration, writes every later value back through the primitive, and mirrors another tab's write in. The marker rides the bindings the way the seed rides the references, so a persisted slice is restored on the screens that bind it.

Four promises and no more. It is best-effort: a value over the slice's cap is not written, the page hears `gofastr:persist-overflow`, and nothing else changes, so an app can say "you have run out of room" rather than lose a write in silence. It is namespaced by the primitive, not by the slice. It is size-bounded, because a signal is re-serialised on every change; megabyte-scale data wants the primitive or the layer above it. And it is invisible to the server: no cookie, no header, no post. The restore is asynchronous by construction, so first paint always shows the server's seed; a screen that must know a browser-held value at first paint still wants the cookie mirror `ui.Banner` uses.

`Persist` implies `.Global()`, so the app-global merge rule keeps a partial render from clobbering the browser's value, and persisting a computed slice panics: it is recomputed from its dependencies on every load, so a stored copy would restore a stale answer.

A persisted signal keeps its trust. Persistence laundered `runtime.js`'s `untrusted` flag into stored markup: the write-back stored an untrusted value like any other, and the restore called `setSignal` with no options, which clears the flag, so a value seeded from `location.search` on one load took the `innerHTML` branch on the next. An untrusted value is never persisted now (reason `untrusted`) and a restore is `{untrusted: true}`. Test: `TestPersistedSliceDoesNotLaunderAnUntrustedValue`, both ends in a real browser; each half fails on its own when its half of the fix is removed.

## The lint arm

`storage-key-raw` asked for "a literal prefix" beside the component encoding. `'x' + encodeURIComponent(v)` satisfied that while the value still reached every key on the origin starting with an `x`, and `encodeURIComponent(v) + '.gofastr'` satisfied it while naming nothing at all. Every storage key in the tree already reads `gofastr.` or `gofastr:`, so that spelling is now the rule, machine-checked: a Web-storage key with a component-encoded operand must lead with a literal, or an identifier provably holding one, that opens the framework's namespace. The new arm reads any component-encoded operand, not only an attribute-borne one, because the key the `local` primitive stores under is an application-chosen parameter the provenance walk cannot see, and that is the value that must not be able to name `gofastr.colorScheme`. The tree was already clean.

## Gates

Run on this branch, on upstream `main` at `f1220e58`. `gofmt -l .`, `go vet ./...`, `go run ./cmd/repolint .`, `make csp-check`, `make embed-check` and `make analyze` are clean. `core-ui/check` runs the lint arm's three mutations on the real module (the namespace emptied, a namespace outside `gofastr.`, the namespace moved behind the value), each failing at its line. `core-ui/runtime` drives the primitive end to end in Chrome (round trip, the fallback for tiny values only, missing keys, namespace containment, prefix enumeration on both engines, subscribe and watch across tabs, plus the four fix tests above) and runs the attrdoc, fragment, budget and lint gates. `core-ui/store` drives the persisted slice in Chrome (survives a reload, converges across two tabs, refuses a value over its cap, keeps its trust, an unpersisted slice is never stored) plus the source gates on registration, `Global`, computed refusal, the cap range and the marker on every binding. Every fix above was watched failing first. There is no JS test runner in this repo, so runtime JavaScript is tested from Go, static lints plus chromedp, which is what every browser test here is.

## Review fixes

The maintainer's pass found six things, all fixed in one commit on this branch, each with a test that was watched failing first. The widened storage-key lint arm had no test that failed when reverted; a fixture with a parameter-rooted key and no literal now fires, and reverting the widening fails it. `persist.js` compared its cap against `text.length`, UTF-16 units, while `PersistMax` is declared in bytes; it measures UTF-8 now and reports bytes on the overflow event, proven with a CJK value that is under the cap in units and over it in bytes. The overflow reason enum in three docs lacked `untrusted`; it is there, and a test parses each document's enumeration so it cannot drift again. `remove()` announced on BroadcastChannel even when the native storage event already carried the removal, so a subscriber heard one remove twice; it follows `set()`'s rule now, proven across two tabs. The fallback's `set()` reported every throw as `quota`; a `SecurityError` is `unavailable` now, matching what `get`, `keys` and `remove` already said. And the persisted-slice restore had no newer-value guard, so a value typed while IndexedDB was still answering was overwritten by the previous session's; it applies only while the signal still holds what it held when the read was requested, proven with a read that resolves late.

## Docs

`core-ui/ARCHITECTURE.md` and `framework/docs/content/runtime-contract.md` carry the primitive's API and its marker-less loading; `signal-store.md` gains the persisted-slice section; `ui-capability-map.md` gains the row, and it says local-first, not offline-first. The changelog entry is on this branch.

## Contract changes for callers

None. `Slice.Persist()` and `PersistMax(n)` are new methods, `__gofastr.local` is a new marker-less module, and the `storage-key-raw` lint arm found nothing in the tree to fix. The one internal change is that the runtime's five Web-storage keys now go through the primitive under the `gofastr.state.` namespace they already used.

## Left alone, on purpose

No schema, no version, no per-collection cap: `core-ui` takes no position on what a stored value means, and that is the stacked PR. The fallback stays a fallback, capped at 8 KiB per value and adopted away the first time IndexedDB opens. `get` and the enumerations' values are not validated; the primitive round-trips JSON and a caller that wants a type checks it. No sync, no queue, no conflict resolution.


DonaldMurillo (changes requested):
Reviewed on `7ad7224`. Four changes requested, all small, one of them a guard with no test behind it.

CI has not run on this PR — the only check present is CodeRabbit, reporting "Review rate limited", which means it read nothing. Everything below I ran locally.

## What I verified

- `go build ./...`, `go vet ./core-ui/...`, `make analyze`, `gofmt -l`, `make csp-check`, `make embed-check`, `make repo-lint`: all clean.
- `go test` on `core-ui/store`, `core-ui/check`, `core-ui/widget`, `framework/docs`: pass.
- The browser suites, run one at a time: `core-ui/store` (5 scenarios, 4.4s) and `core-ui/runtime` in full (384s, including the attrdoc and budget gates). Pass.
- semlint `-sets=all` over the diff, including the browser-storage rule set: two hits on `core-ui/runtime/src/local.js`, both false positives. The read/write key-mismatch hit is wrong because `strip` and `storedPrefix` are the same builder as the write side; the unbounded-growth hit is bounded by the collection caps a layer up.
- I traced every failure mode of `local.js` the "every call settles, none throws" contract has to cover — `indexedDB` absent, `open` throwing synchronously, `onerror`/`onblocked`, a transaction aborting mid-flight, the DB deleted underneath, `localStorage` throwing `SecurityError`, a malformed escape during enumeration — and found no path that rejects or fails to settle. The adoption path is one-way and a record cannot survive in both engines past the first IndexedDB session.
- `LS_MAX_BYTES` comparing UTF-16 code units is documented in the code as deliberate ("the fallback engine's own unit"), so that one is fine. The persist-side version is not — see below.

## 1. The widened `storagekeyraw` arm has no test that fails when you break it

`core-ui/check/runtimeshapes.go:3451`. The PR widens the no-namespace arm from `wrapped != "" && !hasLiteral` to `!hasLiteral`, and the rule's own comment (`runtimeshapes.go:3233-3239`) says the point is to ask about **any** component-encoded operand rather than only an attribute-borne one, "which is exactly the value that must not be able to name `gofastr.colorScheme`".

I reverted that one condition to its pre-widening shape and ran the whole `core-ui/check` package plus the repo-wide `TestStorageKeysEncodeAttrValues`:

```
ok  github.com/DonaldMurillo/gofastr/core-ui/check    2.068s
ok  github.com/DonaldMurillo/gofastr/core-ui/runtime  0.269s
```

Nothing fails. The `local.js` mutations in `TestStorageKeyRawAcceptsTheLocalPrimitive` never reach this arm — an emptied or foreign `PREFIX` literal still counts as a literal, so those cases fall through to the foreign-namespace arm instead. The shape the widening exists for, `localStorage.setItem(encodeURIComponent(param), v)` with no literal at all and non-attribute provenance, is never exercised.

This is hard rule 11 on a security lint. One fixture with a parameter key and no literal, expected to fire, closes it.

The rest of the lint is well covered, for the record: I broke `gofastrNamespace` and the lead `nsIdx` check and both produced real failures, including the vacuity assertions.

## 2. `persist.js` enforces a byte cap in UTF-16 code units

`core-ui/store/persist.js:80` gates on `text.length > entry.max`. `core-ui/store/persist.go:115` documents `PersistMax` as "an explicit cap **in bytes** of JSON-encoded value", and `PersistDefaultMaxBytes = 64 << 10`. Both `framework/docs/content/signal-store.md` and the `data-fui-signal-persist` row in `runtime-contract.md` repeat "the slice's cap in bytes".

`String.length` counts UTF-16 code units. A 1,500-character CJK string is 4,502 UTF-8 bytes and 1,500 code units, so it passes a `PersistMax(2048)` check and is stored at 2.2× the declared cap. The `gofastr:persist-overflow` event then reports `size` in code units and `max` in bytes, so a page cannot even compare them.

`core-ui/runtime/src/local.js:200-202` already ships the right helper — `bytesOf`, using `TextEncoder` — and its comment calls it "the unit a layer above budgets in". `persist.js` just does not call it. `persist_e2e_test.go:226` pins the cap with ASCII-only values, which is why nothing caught it.

## 3. The persist-overflow reason enum is documented with four values and emits five

`core-ui/store/persist.js:69` refuses an untrusted signal with `notify(name, 'untrusted', …)`. Three places list the reasons and all three say four:

- `framework/docs/content/runtime-contract.md:109` — `reason: "size"|"quota"|"encode"|"unavailable"`
- `core-ui/ARCHITECTURE.md:126` — the same string
- `framework/docs/content/signal-store.md` — "reason one of `size`, `quota`, `encode`, `unavailable`"

The CHANGELOG does mention `untrusted`, so the value was not forgotten; the two tables and the architecture row drifted inside the commit that introduced it. A page switching on the documented enum drops the untrusted refusal silently, which is the one refusal that means "something tried to launder an attacker-supplied value into storage".

## 4. `remove()` delivers one removal twice on the IndexedDB path

`core-ui/runtime/src/local.js:333-352`. `remove` calls `localStorage.removeItem(...)` unconditionally and then `announce(key)` when the IndexedDB delete succeeds. When localStorage actually holds that namespaced key, other contexts get the native `storage` event **and** the BroadcastChannel message, so `notify(key)` runs twice and every subscriber re-reads twice.

`set()` has a comment three lines up explaining why not to do this: *"No announce: a localStorage write already reaches the other tabs of this origin through the native storage event… Announcing too would deliver the same change twice and fire every subscriber twice."* The fallback arm of `remove` honours it too ("No announce on the fallback"). Only the IndexedDB arm does not, and the CHANGELOG promises "never both, so one write delivers once".

The state it needs — a localStorage entry still present under `gofastr.state.` while IndexedDB is the engine — is the one `local.js:254` says the `storage` listener exists for: "the fallback engine and any writer that is not this module". Blast radius is duplicate callbacks, not data loss.

## 5. `set()` reports a blocked origin as a full quota

`core-ui/runtime/src/local.js:313-317`:

```js
try {
  window.localStorage.setItem(PREFIX + encodeURIComponent(key), text);
} catch (_) {
  return { ok: false, reason: 'quota' };
}
```

Every throw becomes `'quota'`. In a browser with cookies and site data blocked, the accessor itself throws `SecurityError`, and the module already has the right word for that — `'unavailable'` — which `get`, `keys`, `entries` and `remove` all return correctly on the same page.

Driven in Chrome, one page with both accessors throwing:

```
available → {idb:false, ls:false, engine:'none'}
get       → undefined
keys/entries/remove → {ok:false, reason:'unavailable'}     ← right
set       → {ok:false, reason:'quota'}                      ← wrong
```

and a counter-probe on the real quota path (origin's localStorage filled to a genuine `QuotaExceededError`, then one 60 KB key freed) gives `'quota'` while full and `{ok:true}` after freeing, so the reason is truthful there and wrong only for denial. An app that maps `reason === 'quota'` to "storage is full, free up space" tells a user with storage blocked to delete things. Branch on `err.name`.

## 6. The persisted-slice restore has no newer-value guard

`core-ui/store/persist.js:103-120`. The write-back direction has the `echoing`/`entry.last` guards; the restore direction has none:

```js
// Seed from the browser. SSR painted the server's value; this is
// the one this browser last held, so it wins.
NS.local.get(name).then(apply);
```

"It wins" assumes nothing changed between first paint and the read landing. If the user types before IndexedDB answers — a cold store, a slow disk — the write-back stores the typed value first, and then `apply` overwrites the signal with the value the `get` read *before* the typing.

Driven in Chrome with the read delayed 400 ms and a `setSignal` at +100 ms:

```
signal = "stored yesterday"    while    record = "typed now"
```

The UI shows the old value over a record the user just changed, and the two disagree until the next edit. Applying the restore only when the signal still holds the SSR-painted default closes it.

`framework/local/local-bridge.js:192-229` has the same shape on the seed bridge; raised on #433.

AlexCiccolella125:
All six fixed in `595bb994`, each with a test that was watched failing first, plus a style commit over the branch's comments and copy.

1. `TestStorageKeyRawFiresOnUnnamedParamKey`: a parameter-rooted key with no literal fires the no-namespace arm; reverting the widening fails it with the foreign-namespace message instead.
2. `persist.js` measures the JSON text's UTF-8 length and reports bytes on the overflow event. `TestPersistedSliceCountsItsCapInUTF8Bytes` uses 30 CJK characters, 32 units under a 64 cap and 92 bytes over it.
3. `untrusted` is in the enum in all three docs, and `TestPersistOverflowReasonsAreDocumented` parses each document's enumeration against the reasons `persist.js` emits, so the three cannot drift apart again.
4. `remove()` announces only when localStorage did not hold the key, `set()`'s rule. `TestLocalRemoveDeliversOnceWhenBothEnginesHeldTheKey` plants the key in both engines and counts one callback in the other tab.
5. The fallback's `set()` branches on `err.name`: `QuotaExceededError` and Firefox's `NS_ERROR_DOM_QUOTA_REACHED` are `quota`, anything else is `unavailable`. `TestLocalFallbackNamesABlockedStoreUnavailable` stubs each name.
6. The restore applies only while the signal still holds what it held when the read was requested. `TestPersistedSliceRestoreDoesNotOverwriteANewerValue` reads at request time and resolves 500 ms later with a `setSignal` at 100 ms, the shape of your probe; the first version of that test delayed the read itself and could not fail, so it was rewritten until the mutation failed it.

Budget after the changes: `persist.js` 1011 gzip bytes, `local.js` 2378, against the 3072 goal. The style commit is comment and copy only; a go/parser comment-strip and a minifier comparison against the previous commit show no code token changed, and the four string literals that changed each swapped one dash for a semicolon in a message a test does not pin.

core-ui/runtime/src/local.js
DonaldMurillo, line 1: We def need to d-slopify this file. I also don’t know if this is the right terminology.
- lets not use “state” as an entry point name I would call it “localState”
- I also thinkwe should have a indexdb dedicated api and a local storage dedicated api
- I also think it should be flexible enough to work hand to hand with something like dexie (but please do not bring any node modules use jsdeliver or something similar for testing)

PR #433, opened 2026-09-18
feat(local): local-first state declared in Go, kept in the browser

AlexCiccolella125:
Stacked on DonaldMurillo/gofastr#432, the `local` primitive and `store.Persist`; its commits sit in this diff until it merges. DonaldMurillo/gofastr#431, the runtime `?v=` cache fix, is also on this branch and was cherry-picked out first.

## Why

A server-rendered app that keeps something large and user-owned in the browser (a team builder's rosters, a draft, a preference the page needs at first paint) has had no supported way to do it. Every such app wrote its own island with a document script over raw IndexedDB or `localStorage`: an invented key namespace, no size cap, no schema version, and a hand-rolled way to get the value into a request. This is the opinionated layer on the primitive DonaldMurillo/gofastr#432 adds, and the four bridges a Go screen needs to read or write what the browser keeps. Local-first, not offline-first: no queue, no conflict resolution, nothing reconciling in the background, and `ui-capability-map.md` says so in its new row.

## framework/local

A `local.Store` is declared once per app with named collections. Each `local.Define[T]` is a Go record type with a JSON round-trip, an optional key field, a size cap per record and per collection, and a schema version with `Rename`, `Default`, `Remove` and `Func` migrations the browser runs once, under `navigator.locks` when present. Caps default to 64 KiB, 1000 records and 1 MiB per collection with ceilings the declaration cannot pass; a `Mirror` collection is clamped to cookie size, and a store's mirrored collections share one 4 KiB budget (`MirrorStoreMaxBytes`, a panic at `Define`, refused in the browser with reason `mirror`) so they cannot cook a Cookie header into a 431. The browser API is generated from the declaration: `__gofastr.localStore(app).collection(name)` is `get`, `put`, `delete`, `list` with ordering, `count`, `subscribe`, `clear`; every call settles with `{ok, reason}`, a refusal raises `gofastr:local-error` with the reason and the cap, and a collection whose migration did not complete is gated rather than served on a schema this build cannot read.

Two constraints shaped the design. The declaration cannot be generated into the module: the lint enumerator refuses a `RegisterBehavior` whose source is not an embedded variable, and the hard-rule-5 gate reads marker literals. So the modules are static and linted like every `src/*.js`, and the declaration reaches the browser through the host's existing extra-script rail, the one computed reducers use. That is also the safer shape: a declaration carried on DOM markers would let markup planted in an island response redefine a collection's caps. A marker names a collection; the manifest says what it is. `Store.Script()` serves it as one expression, the rail URL and the mount step, because doing one half without the other 404s the manifest silently. And there is no response builder or action envelope for app handlers to extend: effects ride response headers `rpc.js` reads, and app handlers decode bodies strictly. So the download bridge is a header, `X-Gofastr-Local`, accumulated like the toast header, and the upload bridge is a reserved body field the wrapper strips before the app handler decodes.

## Three modules through the seam

Registered beside the Go that renders their markers, the way `framework/headless` registers its module, and split at responsibility lines: `local-store` (`Requires("local")`, marker `[data-local-store]`) keeps records and enforces the caps; `local-bridge` (`Requires("local-store")`, marker `[data-local-seed]`) is every way the store reaches a Go handler, the seed, the mirror cookie, the upload and the download, and a page whose store keeps its records to itself never loads it; `local-migrate` (`Requires("local-store")`, `LoadIdle`) is the version steps, asked for by name the moment a rewrite is due, so a collection with no version step never runs a line of it. Each is under the per-module budget DonaldMurillo/gofastr#423 introduced, and each sets its loaded flag on its second line.

## Four bridges to a Go screen, nothing in the background

- **Seed**: `local.SeedSignal(coll, key, slice)` fills a `core-ui/store` slice from a record after hydration, writes every later signal value back, and re-applies on every scan so a route-cache replay repaints. It implies `Global()` and refuses a `Persist`-ed slice.
- **Mirror**: a `Mirror` collection also writes each record into a cookie, `Secure` on https, so a render reads it with `local.Get` at first paint. The server accepts only a declared and mirrored collection, a valid key, and a value under the cap that is valid JSON and unmarshals into `T`.
- **Upload**: `local.Send(items...)` declares the collections or keys that ride an RPC request as the reserved field `__local`. It fails closed: a request whose declared records could not be attached, or that is past the bound `Send` derives from the caps it names, is not sent at all. `Upload.Wrap` bounds the body, parses strictly, refuses anything the `Send` did not name (400), enforces the caps (413), strips the field, and puts the records on the context for `local.Get` and `local.List`.
- **Download**: `local.Put`, `local.Delete` and `local.Clear` write records back through the response header, pure ASCII, 16 KiB cap; every op runs through the same `put` a page write uses, so caps and subscribers apply. `local.ClearOnNextLoad` covers a full-navigation logout.

Every read says which source answered. `local.SourceUpload` arrived on a request whose trigger declared it and whose handler is wrapped by the `Upload` that declared it; `local.SourceMirror` is a client hint any script on the origin can write and any client can forge. A handler that treats them alike has an unauthenticated write channel it did not mean to open, which is why `Source` is a return value and not a doc note. A `Get` or `List` on a request nobody wrapped warns under `GOFASTR_DEV`, because a non-mirrored collection can only arrive through the wrapper and the missing line is one the compiler cannot ask for.

## One new runtime attribute, owned by rpc

`data-fui-rpc-with="<module[,module]>"`: the trigger names the modules the runtime loads before dispatching, and the request hooks on `__gofastr._rpcHooks.request` decorate the request the fetch is built from; response hooks run on a 2xx after the runtime's own headers. A hook can mark the request fatal, and a named module that will not load does so: the dispatch is cancelled and the page hears `gofastr:rpc-refused`, because a trigger that names a module is declaring a precondition. Without it the upload hook would race the marker scan on a click that lands before the module loaded. It is `rpc`'s own seam, in the attribute table, the runtime contract and `fragments.go`, proven in `core-ui/runtime/rpc_hooks_e2e_test.go`.

## What was cut before opening this

The branch carried a week of additions built for an out-of-tree consumer, none with a caller here: `Adopt` (a version step over a foreign `localStorage` key), `AnyKey` (one record whose key the page names at click time), `SeedCount`, `Store.Serve` beside `Script`, `Collection.Caps()`, `SeededSignal.Name()`, a `FromContext` escape hatch, browser `list` options nobody read, and a doc twice this length. They are out (two `refactor(local)` commits at the top of the branch), with the tests and doc sections that described them, and they return with the consumer that needs them. The example that motivated the package, a team builder whose roster survives quitting the browser, lives outside this repository at AlexCiccolella125/gofastr-team-builder and is linked from the examples catalog; `examples/site` at `/forms/draft-notes` exercises all four bridges on one page.

## Review fixes

The maintainer's pass found four Majors and eleven more; all are fixed on the branch, each Major and most of the rest with a test that was watched failing first.

The four Majors. A tab open across a deploy drops its cached readiness when another tab writes the collection's version entry, so its next put re-reads the stored version, sees one above its own, and refuses; before, the cached readiness let an old-schema record land in a migrated collection forever (`TestE2E_ATabOpenAcrossADeployDoesNotWriteTheOldSchema`). `clear()` and `delete()` queue on the same per-collection chain as `put`, so a put cannot commit after the enumeration a logout never saw; four rounds of a put and a clear fired in one tick leave zero records (`TestE2E_AClearIsNotOutrunByAPut`). `asciiJSON` escapes DEL, which `encoding/json` leaves raw and HTTP/2 drops the whole header for; the header test carries a DEL and a C0 byte now. The upload pre-flight counts UTF-8 bytes, the unit `MaxBytesReader` counts; 120 CJK characters against a 260-byte bound are refused before the fetch (`TestE2E_TheUploadPreFlightCountsBytesNotUnits`).

The rest. The mirror read path stops at `MaxRecords` and `MaxBytes` per collection, so fifty planted cookies against a cap of four read as four. `list()` and `count()` report a failed enumeration with reason `unavailable` instead of resolving empty. The mirror cookie budget is summed across every declared store in the process, and the browser counts every store's cookies on the origin. `Send(c, c.Key("a"))` panics at declaration with both items named. `rpc.js` awaits response hooks inside the try, and the contract says a hook must not read the body. `sendItem.bound` multiplies in `int64` and clamps before narrowing. The mirror cookie's JSON is validated strictly, so a duplicate or case-folded key is ignored the way the upload refuses it; the decode into `T` stays lax on both paths, because the browser may keep more than the Go type reads. The seed refuses a second element that binds one signal to another record, and a restore that lands after the user typed keeps the typed value. Two tabs writing at once are serialised by a Web Lock on the collection, with the per-tab caveat for Safari in the doc; a ten-round synchronised burst from two tabs stays under the cap, and passes the cap in round one with the lock removed. The doc says the auth battery's logout does not clear the store and names the two wiring points, and `examples/site` gained a logout route that calls `ClearOnNextLoad`. `framework/ARCHITECTURE.md` lists the package with its imports. The browser's FormData upload arm has a multipart scenario; the two comments that pointed at the extracted example name its repository; the release manifest counts eleven shards; the two changelog entries on unreleased behaviour are folded into their `Added` entries.

Also on the branch, at the maintainer's request: the `unslop` writing skill vendored under `.claude/skills/`, and a style commit over the package's comments and copy with a go/parser comment-strip and a minifier comparison showing no code token changed.

## Gates

`gofmt`, `go vet`, `repolint`, `csp-check`, `embed-check`, the repo vettool and the JS lints in `core-ui/check` over the tree with the registered modules included are clean. `framework/local` passes its unit tests and twenty-seven browser scenarios (record survives reload and a second browser is isolated; caps refuse with a reason and raise the event; `list` orders; a v1→v2 migration runs once and stamps, a failed one neither stamps nor answers, a downgrade is refused; the mirror carries the migrated record, is `Secure` on https and refuses to overfill the header; the seeded signal survives soft navigation and the route cache; the upload delivers only the declared records and refuses rather than send without them; a refused download op is reported; a second tab hears the write; `ClearOnNextLoad` with and without a store marker; racing puts cannot pass the cap; a store with no manifest says which URL it wanted). `core-ui/runtime` in full (attrdoc, fragment, budget and lint gates plus every chromedp module test), `framework/docs` (link, compile-snippet, security-rule and common-mistakes gates over both docs), `examples/site` including the draft-notes suite and the axe scan, and `TestReleaseManifestMatchesCI` pass.

Proven by breaking, then restored: the unmirrored-cookie refusal and the undeclared-key refusal each fail their test when removed; a mutated copy of the store module with a raw attribute-borne key or a foreign namespace, and of the bridge with an unencoded cookie key, fails the corresponding lint at its line; a migration that stamps before its writes settle fails the failed-migration scenario.

## CI

A new shard `browser e2e · framework-local` runs the package isolated, the same Chrome-contention class as `framework-headless`; it leaves the main job's `-p 2` run. `scripts/release-required-checks.txt` lists it. Branch protection needs the new context added once the check has run.

## Docs

`framework/docs/content/local-state.md` is the contract: declare, serve, the browser API, the four bridges, the rules (best-effort by construction, every call settles and says why, no secrets in the store and why, no server-side memory of browser state, multi-tab through the primitive's `watch`, first-paint needs go through the mirror). The capability map gains its row, `core-ui/ARCHITECTURE.md` and `runtime-contract.md` carry the new attribute and the three modules, `framework/local/agents.md` says how the package is put together with a compile-gated snippet, and the changelog names all of it.

## Where this bends a rule, and why

- The mirror cookie is unsigned. A cookie the browser writes cannot carry a server signature: the server never sees the value before the browser stores it, and the module has no key (`banner.go`'s cookie is unsigned for the same reason). Integrity would mean a server-minted value the browser only echoes, which is server-owned state, not built here. `Source` is how a handler tells the two apart.
- The namespace nests under the primitive's (`gofastr.state.local.<app>.<collection>`), not beside it. A second top-level namespace would put a second literal in the primitive or have the framework touch Web storage directly, and both put opinion where it must not go. The cookie names, which are the framework's own sinks, use `gofastr.local.`.
- `gofastr.contracts.yml` is unchanged. It relaxes rules; nothing here needs one. The one `//gofastr:allow(GOFASTR1404)` sits on the clear-on-next-load cookie, deliberately not `HttpOnly`, with its reason on the line.

## Left alone, on purpose

Offline-first sync, conflict resolution and a background queue stay a non-goal; the capability row says so. The dev-mode warning for a read outside `Upload.Wrap` is heavier than a type could be, and stays because the failure it catches is silent. There is no JS test runner in the repo, so runtime JavaScript is tested from Go (static lints plus chromedp), which is what every browser test here is.


DonaldMurillo (changes requested):
Reviewed on `81d4333`. This is careful, well-tested work — the gates back it up and the design decisions are the right ones. Four findings need fixing before it lands. Two of them lose user data silently, and both were reproduced deterministically in a browser.

CI has never run on this branch. The only check is CodeRabbit reporting "Review rate limited", which means nothing has read these 8,551 lines. Everything below I ran locally, and every claim carries the probe that produced it.

## Gates, run locally on this branch

| gate | result |
|---|---|
| `go build ./...` | clean |
| `make analyze` (the whole vettool set) | clean |
| `gofmt -l`, `make csp-check`, `make embed-check`, `make repo-lint` | clean |
| `go test ./framework/local/` — 19 browser scenarios | pass, 10.4s |
| `go test ./core-ui/store/` — 5 browser scenarios | pass, 4.4s |
| `go test ./core-ui/runtime/` — attrdoc, fragment, symbol, budget, lint gates | pass, 384s |
| `go test ./core-ui/widget/ ./core-ui/check/ ./framework/docs/` | pass |
| `go test ./examples/site/` — dogfood plus the axe scan | pass, 748s |
| `TestReleaseManifestMatchesCI` | pass |
| semlint `-sets=all`, 1,471 sites over the diff | no true positives |

The CI wiring is right, which is the part this repo has been bitten by before: `framework/local` excluded from the main `-p 2` job with the reason in the comment, its own isolated shard, the shard name generated by the job template so it matches `release-required-checks.txt` exactly. `data-fui-rpc-with` is in the `core-ui/ARCHITECTURE.md` attribute table with the fail-closed semantics spelled out, so hard rule 5 holds. `screen_localnotes.go` ships zero CSS and zero hand-rolled structure, composing `ui.PageHeader/Stack/Section/Card/Form/FormField/Cluster/Button/Tag` — and `actOnDraft(src) { return src == local.SourceUpload }` is exactly what the flagship demo should be teaching. Every numeric claim I spot-checked in `local-state.md` matches its constant.

Beyond the suites, I drove Chrome directly at the failure paths the tests do not reach, and most of them hold. Web Locks is genuinely held during a migration — `navigator.locks.query` shows the exclusive lock on tab A's clientId with tab B's request queued, and tab B's `get` stays pending rather than reading half-migrated data. A migration failing on record 5 of 10 leaves the collection gated with five `gofastr:local-error` events and no stamp. The `navigator.locks`-free path (the Safari shape, both tabs migrating in the same instant) stamps twice but loses and duplicates nothing. IndexedDB quota, forced through CDP, settles `{ok:false, reason:'quota'}` and the next call still works. Download ops apply whole or not at all across an aborted fetch. Five away-and-back navigations with the route cache warm: the record wins on every leg, the listener count stays at one, and one later `setSignal` produces exactly one write — no seed leak in either direction. `clear()` with an unremovable entry reports the failure and leaves `count()` and `list()` agreeing.

I re-broke the three guards the description claims were proven by breaking — the undeclared-collection refusal, the response-header cap, the mirror-cookie validation — and each failed its test with the guard gutted. That claim holds.

---

# Major

## 1. A tab open across a deploy writes old-schema records into a migrated collection, permanently

`framework/local/local-store.js:239-244` and `:163-168`.

`whenReady` memoises the migration once per page and never invalidates it:

```js
const whenReady = (coll) => {
  const spec = specOf(coll);
  if (!spec) return Promise.resolve(false);
  if (!ready.has(coll)) ready.set(coll, migrate(coll, spec));
  return ready.get(coll);
};
```

and the cross-tab watcher deliberately drops the one event that would invalidate it:

```js
P.watch('local.' + app + '.', (k) => {
  const rest = k.slice(('local.' + app + '.').length);
  const colon = rest.indexOf(':');
  if (colon < 0) return;            // a version entry, not a record
  notify(rest.slice(0, colon), rest.slice(colon + 1), 'tab');
});
```

So:

1. A user has a tab open on build v1. Stored version is 1, `ready` caches `true`.
2. You deploy v2 with `Rename("body", "text")`.
3. The user opens a second tab. v2 migrates every record and stamps version 2.
4. The user returns to the first tab and types. `put` → `gate` → the cached `true` → a `body`-shaped record lands in a collection whose stored version is 2.
5. Nothing ever migrates it. Every later load has `stored == declared == 2`, so `migrate` returns at `if (have === target) return true` before looking at a single record. Every v2 reader sees `text === undefined` on that one record, forever.

The guard for the mirror image of this already exists, eight lines up:

```js
// A stored version ABOVE the declared one is a rollback: this
// build's steps cannot undo what a newer build wrote, and
// reading those records as if they were this schema is how a
// deploy that gets rolled back corrupts the browsers that
// already moved on. Refuse the collection and say so.
if (have > target) { fail(app, coll, '', 'version', have, target); return false; }
```

The v1 tab in step 4 is in exactly the `have > target` state that comment describes. It just never re-evaluates, because the answer was cached at page load. The read side is guarded and the write side is not, for that one reason.

`TestE2E_SecondTabSeesTheWrite` covers cross-tab notification at one version; `TestE2E_AVersionDowngradeIsRefusedNotAccepted` covers the read side on a fresh load. Neither reaches this.

Two shapes for a fix: route version-entry writes through the watcher so `ready` resets, or re-read the meta entry inside the collection's serialised put chain — `put` already serialises, so the cost lands only on writes.

## 2. `clear()` and `delete()` bypass the collection's put chain, so a logout reports success over records that survive it

`framework/local/local-store.js:273-278`, `:324-332`, `:378-394`.

`serial(` has exactly one call site in the file — inside `put`. The comment that justifies the chain names only the case it was built for:

> *Two puts racing each other both read the same total, both decide they fit and both land … it is the common case in a page that writes on every keystroke.*

A put racing a **clear** has the same read-then-write shape and a worse outcome: `clear()` enumerates keys in one transaction, the queued put commits after it, and the key the enumeration never saw is never removed. `clear()` then resolves `{ok:true}`.

Driven in Chrome against a mirrored collection, put and clear fired in one tick:

```
same tab,  8 rounds: put-survived-own-clear=8  clearReportedOK=8  mirrorCookieSurvived=8
cross tab, 8 rounds: tabA-put survived tabB-clear=8  clearReportedOK=8
delete variant:      delOk:true  exists:true
```

Deterministic, 8 out of 8, and the mirror cookie survives with the record — so the previous user's value is still in the browser *and* still riding the `Cookie` header to the server on the next request. The suite states the stake itself: *"the previous user's records are gone is the only thing a logout is for"* (`local_e2e_test.go:570`).

Route `clear()` and `delete()` through the same per-collection `serial` chain.

## 3. `X-Gofastr-Local` can carry a raw DEL byte, and HTTP/2 drops the whole header

`framework/local/response.go:109` (`asciiJSON`).

`asciiJSON` escapes every rune at or above `utf8.RuneSelf` and leaves the rest to `encoding/json`. `encoding/json` does not escape **DEL (0x7F)**. A DEL anywhere inside a `Put` value reaches the header value raw. Proven against a real server:

```
header  = {"app":"delprobe","ops":[{"c":"vals","k":"k","v":{"text":"a\x7fb"}}]}
          → non-printable byte at offset 59: 0x7f

wire    = X-Gofastr-Local: {"app":"delwire", ... "a\x7fb" ...}      (raw socket read)

http.Get(srv.URL)
  → net/http: HTTP/1.x transport connection broken:
    malformed MIME header line: "X-Gofastr-Local: ...a\x7fb..."
```

The client gets no response at all — the op and the body go together.

What makes this more than an RFC footnote: `framework/ui/toast.go:63` — the header this one is explicitly modelled on ("Calls accumulate on one response the way `ui.AddToast` does") — already carries the fix and the reason:

> *encoding/json does not escape DEL, so a `\x7f`-bearing title lands as a raw byte in the header value, and Go's HTTP/2 writer silently drops an invalid header — losing EVERY toast in the response, not just the poisoned one.*

`core/textsafe.HasControlBytes` exists for this and tests `s[i] < 0x20 || s[i] == 0x7f` by name. Under HTTP/2, which is what a deployed app serves, the header vanishes with no error — so **every download op is silently lost**, on a bridge whose only job is writing records back. That is worse than the HTTP/1 failure I could reproduce.

`TestPutDeleteClearAccumulateOneASCIIHeader` does assert every byte lands in `[0x20,0x7e]`, so the assertion is right; its sample value is `"héllo 🙂"`, which is above 0x80 and therefore already escaped, so the input never reaches the gap. Worth a DEL and a C0 byte in that test alongside the fix.

One line: send `r == 0x7F` through `writeU` as well.

## 4. The upload pre-flight counts UTF-16 code units against a byte bound

`framework/local/local-bridge.js:297` — `if (max > 0 && text.length > max)`, where `text = JSON.stringify(payload)`.

`text.length` is UTF-16 code units. `max` is a **byte** count the Go side derived and served verbatim on the trigger (`send.go:150`), enforced server-side as bytes of the whole body by `MaxBytesReader` (`send.go:322`). `JSON.stringify` does not escape non-ASCII. Measured:

```
UTF-16 units: 3125   UTF-8 bytes: 8125   ratio: 2.60
a pre-flight max of 6000 would:        PASS
the server MaxBytesReader(6000) would: 413
```

So an app calls `Upload.Max(6000)` — the documented remedy for a trigger carrying a body of its own (`send.go:44-48`) — the user writes a CJK draft, the pre-flight waves it through and the server answers a bare 413. That is precisely "a bare 413 the page cannot see", the failure `send.go:135` and `local-state.md:302-308` both say the pre-flight exists to prevent.

`local-store.js:83` already counts the right unit for keys and records via `bytesOf`/`TextEncoder`. The e2e (`local_e2e_test.go:884`) forces `data-local-max='8'` with ASCII records, where units and bytes are equal, so it cannot see this.

(The other discrepancy — the browser counts only the `__local` payload while the server counts the whole body — is the deliberate `UploadBodySlack` trade-off, not a bug.)

---

# The rest

## 5. The mirror read path enforces the per-record cap and nothing else

`framework/local/read.go:110-139` and `:79-109`. `decodeCookie` checks the collection is declared and mirrored, the key is valid, and `len(text) > def.maxRecord`. Nothing counts records per collection or sums against `def.maxBytes`. Those caps live at `Define` and inside the browser module, and a client that ignores both — curl, or any script on the origin writing `document.cookie` — is simply believed:

```
planted 50 mirror cookies,  declared MaxRecords=4 -> List returned 50 records  (err=<nil>)
planted 300 mirror cookies, declared MaxRecords=4 -> List returned 300 records (err=<nil>)
declared aggregate budget=4096 bytes; 8 cookies carrying 7400 payload bytes -> all accepted
```

The asymmetry is the point: `Upload.parse` enforces both the record count (`limitFor`) and `def.maxBytes` server-side and answers 413. Same declaration, same `T`, two bridges, and only one of them holds the caps. A screen that renders `local.List` or loops over it now has a client-driven bound where the declaration promised four.

Counting per collection in `carriedBy` and stopping at `def.maxRecords`/`def.maxBytes` matches what `decodeCookie` already does per record.

## 6. `list()` and `count()` report a failed enumeration as an empty collection

`framework/local/local-store.js:340` and `:360`:

```js
list(opts) {
  ... P.entries(prefixOf(coll)).then((er) => {
    const out = [];
    for (const e of er.entries) out.push(...)      // er.ok never read
...
count() { return gate(coll).then((no) => (no ? 0 : P.keys(prefixOf(coll)).then((r) => r.keys.length))); }
```

The kernel settles a failure as `{ok:false, reason, entries:[]}` rather than throwing, so `list()` resolves `[]`, `count()` resolves `0`, and `gofastr:local-error` never fires. A Safari private-mode read or an aborted transaction renders "no notes" and tells nobody.

`clear()` in the same file had this bug and it was fixed, with the reason written down: *"An enumeration that aborted used to look exactly like an empty collection, so the logout path reported {ok:true} over records — and mirror cookies — that all survived."* `put()` checks `!er.ok` too. `list` and `count` are the two that were missed.

## 7. The mirror cookie budget is per store; the Cookie header is per origin

`framework/local/collection.go:223-231`. The budget loop sums across `s.colls` — this store's collections. Two stores in one process each get their own 4 KiB:

```
two stores  = 8192 bytes of mirror cookies on every request
five stores = 20480
budget      = 4096, PER STORE
```

The panic that does fire, for one store, names the reason the number exists: *"a header block past 8-16 KiB is a 431 the browser cannot recover from"*. `collection.go:38-40` adds that 4 KiB is sized to leave room for "the session cookie, the CSRF cookie and everything else the app puts on the origin" — which holds for one store.

Multi-store is a designed shape, not a misuse: `honourClearBits` and `restampMirrors` both iterate `Object.keys(window.__gofastr_local)`, and `New` keeps a package-level registry keyed by app id. The browser half has the same blind spot — `mirrorUsed` (`local-bridge.js:77`) skips any cookie not named `gofastr.local.<app>.`, so it sees neither a sibling store's cookies nor the session cookie beside them.

The `apps` map at `local.go:104` is already there.

## 8. `Send(c, c.Key("a"))` is accepted and then 400s on every fire

`framework/local/send.go:72` and `local-bridge.js:256-258`. `gather` pushes every record from `c.list()` into `rows` and pushes the named key again from `c.get(key)`:

```
Send(c, c.Key("a"))  → data-local-send = "vals,vals:a"
wire                 → {"vals":[{"k":"a",...},{"k":"a",...}]}
parse()              → local: record "vals"/"a" appears twice   → 400
```

Every fire of that trigger, with a message pointing at the wire rather than the declaration that caused it. Better to refuse at `Send` an item already subsumed by a whole-collection item of the same collection — it fails at declaration time like every other misuse in this package.

## 9. `rpc.js` response hooks run without `await`

`core-ui/runtime/src/rpc.js:300-303`:

```js
for (const hook of hooks.response) {
  try { hook(node, r); }
  catch (err) { console.warn('[gofastr] rpc response hook failed', err); }
}
```

The request side gets this right (`await hook(node, req)` inside the `try`). On the response side an async hook's rejection escapes the `try` and surfaces as an unhandled rejection. This is the core runtime, on the path of every RPC in every app on the framework.

While you are there: the contract that a response hook must not read `r.body` is load-bearing and unstated. A hook calling `await r.json()` consumes the body and the `await r.json()` four lines down throws.

## 10. `sendItem.bound` overflows int on 32-bit

`framework/local/send.go:101`: `n := si.def.maxRecords * si.def.maxRecord`, both factors capped individually (100 000 and 1 MiB) and the product unbounded. `MaxRecords: 2048, MaxRecordBytes: 1 MiB` is exactly 2³¹. On a 32-bit build that wraps negative, `if si.def.maxBytes < n` is then false so the negative survives, and `MaxBytesReader(w, body, ≤0)` refuses every request through that wrapper. At 4096 records the product wraps to 0 and the bound collapses instead.

No 32-bit target in this repo's CI, so this reaches only a host compiling for 386 or 32-bit ARM. `int64` and clamp before narrowing. The mirror-budget product on `collection.go:223` is safe because the Mirror ceilings bound it.

## 11. The mirror cookie decodes lax where the upload decodes strict

`framework/local/read.go:180` and `:199` use `json.Unmarshal`; `send.go:259` uses `handler.UnmarshalStrict`. Both carry the same `T`, both are attacker-controlled, and the package doc is emphatic that the cookie is the *less* trustworthy of the two. A duplicate or case-folded key in the cookie's JSON is accepted and silently resolved last-wins. The generic `T` is why `laxenvelope` cannot see it — this is the shape that analyzer exists for.

## 12. `wireSeed` binds the write-back to the first spec and reads with the latest

`framework/local/local-bridge.js:181-229`. `seeds` is keyed by signal name; the listener and `c.subscribe` install once inside `if (!entry)`, closing over the first element's `coll`/`key`, while the trailing `c.get(key).then(apply)` runs every scan with the current element's key. Two elements sharing a `data-fui-signal` name with different `data-local-seed` specs read one record and write back to another, silently. App misuse, but misuse the framework can refuse — the reserved-name guard three lines above shows the shape.

Related, and proven in Chrome: the restore direction has no newer-value guard either. `apply` runs whenever the IndexedDB read lands, with nothing checking whether the signal has moved since first paint. With the read delayed 400 ms and a `setSignal` at +100 ms:

```
signal = "stored yesterday"    while    record = "typed now"
```

The UI shows the stored value over a record the user just changed, and the two disagree until the next edit. `core-ui/store/persist.js:103-120` has the same shape; raised on #432.


## 13. Two tabs can push a collection past its declared cap

`framework/local/local-store.js:265-315`. The cap check is read-then-write across two IndexedDB transactions, and the serialising chain is per *document*, so it cannot cover two tabs. Driven in Chrome — a collection at `maxRecords: 3, maxBytes: 1024`, two tabs each firing three ~310-byte puts in the same instant:

```
max records stored in a collection capped at maxRecords=3: 4
```

past the record cap and the byte cap together. The same app open in two windows, both autosaving, is all it takes.

Worth saying that the existing `TestE2E_RacingPutsCannotPassTheCollectionCap` is honest — it uses `Promise.all` with no awaits between puts, genuinely races, and the per-document chain does hold it. The cross-tab dimension is the one nothing covers. `navigator.locks` across the read-then-write is already the idiom this file uses for migrations; the alternative is documenting the cap as per-tab.

## 14. No logout path in this repo clears the store, and the docs do not say the host must

`grep -rn "ClearOnNextLoad\|local.Clear(" --include="*.go"` outside `framework/local/` returns nothing. `battery/auth`'s `logoutHandler` revokes sessions and clears session cookies; it never touches `gofastr.local.*`, and it structurally cannot — it does not know the app's stores. Mirror cookies are planted with `max-age=31536000`.

Shared machine: account A logs out through the auth battery, account B logs in on the same profile, and B's first paint reads A's mirrored records via `local.Get`. `local-state.md:349-360` shows `Clear`/`ClearOnNextLoad` as the logout helpers but never states the wiring obligation or its consequence, and "Common mistakes" — the natural home — does not mention it.

One paragraph naming the two wiring points closes it.

## 15. `framework/ARCHITECTURE.md` does not know this package exists

CLAUDE.md opens with "Before adding, moving, or extracting anything under `framework/`, read `framework/ARCHITECTURE.md`", and calls it the contract. This PR adds `framework/local` and does not touch that file; the only six occurrences of "local" in it are "local runtime resource", "local var" and "package-locality".

It belongs in the L3 leaf list beside `headless` — it imports only `core/*`, `core-ui/*` and `framework/agentsinv`, nothing above it — and the intra-L3 edge to `agentsinv` is the kind the table records. `layering_test.go` passes because it enforces direction, not registration, so nothing catches the omission mechanically.

## Smaller

- **The browser's FormData upload arm has no test.** `local-bridge.js:301-303` (`req.body.append('__local', text)`) is documented at `local-state.md:298` and unexercised — the e2e fixture's form has no enctype or file input, so the runtime posts JSON and the handler asserts a JSON body. The *server's* urlencoded and multipart arms are unit-tested at `local_test.go:364`; the browser half is the asymmetric one.
- **Two comments point at `examples/team-builder`, which this PR moved out of the tree.** `screen_localnotes.go:181` and `e2e_local_test.go:167`. A reader grepping for it finds nothing; name the external repo or the README line.
- **`release-required-checks.txt:6` says "the 10 matrix shards"; `ci.yml` now has 11.** The off-by-one predates this PR (main said 9 with 10 shards) and this PR edited that line to 10 while adding the eleventh. The load-bearing half is right — the check name is generated by the job template and `TestReleaseManifestMatchesCI` passes.
- **The changelog's new `Fixed` entry fixes unreleased behaviour.** The module-serve route and loader it corrects were born after the `v0.85.0` tag and sit in `[Unreleased] → Added` at the merge base, so per the repo's rule it folds into that entry. Same for the `Changed` entry on `storage-key-raw`, which strengthens a lint that has shipped nowhere. Honest caveat: the existing `[Unreleased] → Fixed` section already carries entries of the same character, so this follows the file's practice rather than the rule.

---

## On the design

Two things are worth saying plainly, because they are the parts I would not want changed in a rewrite. `Source` as a return value rather than a doc note is the right call, and the demo honouring it (`actOnDraft`) is what makes it stick. And enforcing the mirror budget at `Define` with a panic that explains the 431, rather than discovering it in production, is the kind of guard this framework needs more of — scope bug notwithstanding.

The open question is not in the code. This adds a 4,000-line public subsystem to `framework/`, a docs page, an MCP-visible doc topic and a new runtime attribute, and it links a third-party repository and a Cloud Run URL from `examples/README.md` as its canonical example. That is a maintenance commitment separate from whether the code is right, and it is a maintainer call, not a review finding.

AlexCiccolella125:
All fifteen are on the branch now, force-pushed after a rebase onto the fixed primitive branch (DonaldMurillo/gofastr#432). The PR body carries a "Review fixes" section with the test for each; the short form:

Majors. 1: the cached readiness is dropped when another tab writes the version entry, so the stale tab's next put re-reads the stored version and refuses (`TestE2E_ATabOpenAcrossADeployDoesNotWriteTheOldSchema`; with the `ready.delete` removed it lands the record). 2: `clear()` and `delete()` are on the put chain (`TestE2E_AClearIsNotOutrunByAPut`, four rounds of put and clear in one tick, zero survivors; without the chain the first round fails). 3: `asciiJSON` escapes DEL, and the header test carries a DEL and a C0 byte (fails with the one-line check reverted). 4: the pre-flight counts UTF-8 bytes (`TestE2E_TheUploadPreFlightCountsBytesNotUnits`, 120 CJK characters against a 260-byte bound; fails on `text.length`).

The rest, in your numbering. 5: the mirror read stops at `MaxRecords` and `MaxBytes` per collection, in `Get` and `List` alike. 6: `list` and `count` raise `unavailable` on a failed enumeration. 7: the budget is summed across every store in the process, with the stores named in the panic, and the browser counts every `gofastr.local.*` cookie on the origin except the clear bit. 8: `Send(c, c.Key("a"))` panics at declaration naming both. 9: response hooks are awaited in the try; the contract says a hook must not read the body; a rejecting hook is caught (`TestRPCResponseHookThatRejectsIsCaught`), and the core bundle sits 13 bytes under its goal. 10: `bound` multiplies in `int64` and clamps before narrowing. 11: the cookie's JSON is validated with `UnmarshalStrict`, so a duplicate or case-folded key is ignored the way the upload refuses it; I kept the decode into `T` lax on both paths, because a record the browser keeps may carry fields the Go type does not read, and erroring on those at read time would be a new contract. Say so if you would rather have `DisallowUnknownFields` there too. 12: the seed refuses a second element that binds one signal to another record, and the restore keeps a value typed while the read was in flight (`TestE2E_TheSeedKeepsWhatTheUserTypedWhileTheReadWasInFlight`, a read that resolves late). 13: the write chain holds a Web Lock on the collection, so two tabs cannot pass the cap either; a ten-round synchronised burst from two tabs stays under it and passes it in round one with the lock removed. Safari has no Web Locks, and the doc says the cap holds per tab there. 14: the doc names the two wiring points and says the auth battery cannot clear the store; `examples/site` gained a logout route calling `ClearOnNextLoad`, with a test on the cookie. 15: `framework/ARCHITECTURE.md` lists `local` with its imports and the `local` to `agentsinv` edge.

Smaller: the multipart FormData arm has a browser scenario; the two comments name the external repository; the manifest says eleven shards; the two changelog entries on unreleased behaviour are folded into their `Added` entries.

Two more commits at your request: the `unslop` writing skill vendored under `.claude/skills/` with its scanner, and a style pass over the package's comments and copy, proven token-identical by a go/parser comment-strip and a minifier comparison. Every gate in the body reran green after the rebase. On the maintenance question: that is yours to make, and the examples catalog's link to the external repository and the Cloud Run URL come out on one word if you would rather not carry them.

